PR #70, opened 2017-01-20
Restrain Potential

Nick-Chaos:
Lio may add an extra potential term to Hamiltonian for penalty the distance between specified pairs of atoms. Details about implementaion and uses are in lio/docs/manual/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 .PHONY: all clean hooks
 
 all:
-	cd g2g; make;
+#	cd g2g; make;
 	cd lioamber; make;
 	cd liosolo; make;
 
 clean:
-	cd g2g; make clean;
+#	cd g2g; make clean;
 	cd lioamber; make clean;
 	cd liosolo; make clean;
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,12 +1,12 @@
 .PHONY: all clean hooks
 
 all:
-	cd g2g; make;
+#	cd g2g; make;
```
ramirezfranciscof: This should not be commented, right?

Makefile
```diff
@@ -1,12 +1,12 @@
...
 	cd lioamber; make;
 	cd liosolo; make;
 
 clean:
-	cd g2g; make clean;
+#	cd g2g; make clean;
```
ramirezfranciscof: This should not be commented, right?

Makefile
Nick-Chaos: No, my bad

Makefile
Nick-Chaos: No, my bad

PR #74, opened 2017-01-26
Modifying SCF.f phase 1

Nick-Chaos:
Break main SCF  into subroutines and start modifications for future RMM break. Also remove unused variables. Code tested on intel and gnu compilers with and without cublas.

Please tick if accept pull request. (sorry if forgot a name)
- [x]  fran
- [x]  nano
- [x]  manu
- [x]  uriel
- [x]  fede


fedepedron:
I have the following errors when attempting to compile with gfortran:

SCF.f90:1079:45: call calc_fock_commuts(fock,rho,X,Y,scratch,scratch1,M)
Error: Symbol ‘y’ at (1) has no IMPLICIT type

SCF.f90:1115:51: FP_PFm(i,ndiis)=scratch(j,k)-scratch1(j,k)
Error: Function ‘scratch1’ at (1) has no IMPLICIT type

SCF.f90:1195:17: fock=basechange_gemm(M,fock,x)
Error: Function ‘basechange_gemm’ at (1) has no IMPLICIT type

lioamber/init_lio.f90
```diff
@@ -77,6 +77,9 @@ subroutine lio_defaults()
 !   Distance restrain options
     number_restr   = 0             ;
 
+!   Dbug
```
manuelF: What is Dbug? 'debug'? Use clearer variable names or explain here what it does.

lioamber/int1_new.f90
```diff
@@ -0,0 +1,837 @@
...
+! Pointers
+      M1=1 ! first P
+      M3=M1+MM ! now Pnew
+      M5=M3+MM ! now S, F also uses the same position after S was used
+      M7=M5+MM ! now G
+      M9=M7+MMd ! now Gm
+      M11=M9+MMd ! now H
+!---- Overlap ,Kinetic energy and Nuclear Attraction
+!      matrix elements evaluation
+! Overlap matrix will be kept, kinetic energy and nuclear attraction
```
manuelF: Use consistent indentation

lioamber/liomods/garcha_mod.f
```diff
@@ -151,6 +151,21 @@ module garcha_mod
       INTEGER :: number_restr, number_index
       INTEGER, ALLOCATABLE, DIMENSION(:,:) :: restr_pairs
       INTEGER, ALLOCATABLE, DIMENSION(:) ::  restr_index
-      DOUBLE PRECISION, ALLOCATABLE, DIMENSION(:) :: restr_k, restr_w,
+      DOUBLE PRECISION, ALLOCATABLE, DIMENSION(:) :: restr_k,restr_w,
      > restr_r0
+!-Dbug
+      Logical :: Dbug
```
manuelF: Same as above

lioamber/liomods/garcha_mod.f
```diff
@@ -151,6 +151,21 @@ module garcha_mod
...
+
+!-para ir rompiendo RMM
+	logical :: break_rmm
```
manuelF: What does this mean? Use a clearer description

lioamber/liomods/garcha_mod.f
```diff
@@ -151,6 +151,21 @@ module garcha_mod
...
+!- Futuro reemplazo de RMM
+      real*8, dimension (:), ALLOCATABLE :: Fock_Hcore, Fock_Overlap,
```
manuelF: Can you separate these in a single line each initializaiton? It makes it more obvious the following comments are per variable here.

lioamber/write_output.f90
```diff
@@ -64,9 +64,8 @@ end subroutine write_dipole
 subroutine write_forces(dxyz, natom, offset, uid)
     implicit none
     integer, intent(in) :: uid, natom, offset
-    real*8 , intent(in) :: dxyz(natom, 3)
+    real*8 , intent(in) :: dxyz(natom+1, 3)
```
manuelF: This looks unrelated, why is it here?

lioamber/int1_new.f90
```diff
@@ -0,0 +1,837 @@
...
+ 706  continue
+!
+ 703  continue
+! end nuclear attraction part --------
+!
+ 700  continue
+
+!
+!
+!--- Debugging and tests -----------------------------------------------
```
manuelF: Do not submit commented code:
- It is never run automatically if they are tests
- the code goes stale, meaning the variables linked here will not apply.
- it makes it confusing to see when a variable is reference but never actually compiled, harder to refactor

lioamber/int1_new.f90
```diff
@@ -0,0 +1,837 @@
...
+!
+       if (l1.eq.l2) then
+        f1=sq3
+        d0s=d0s+(s0s(n)-s1s(n))/z2
+        d1s=d1s+(s1s(n)-s2s(n))/z2
+        d2s=d2s+(s2s(n)-s3s(n))/z2
+       endif
+!
+!
+      do 706 l3=1,3
```
manuelF: Nitpick: All these variables could use better names. If we are rewriting these into new code, try to improve the readability of these as well.

Can do in a following PR.

lioamber/int1_new.f90
```diff
@@ -0,0 +1,837 @@
...
+       Fock_Overlap(k)=Fock_Overlap(k)+ovlap*cc
+       Smat(iin,jj)=Smat(iin,jj)+ovlap*cc
+       Smat(jj,iin)=Smat(jj,iin)+ovlap*cc
+       Fock_Hcore(k)=Fock_Hcore(k)+term
+       endif
+!
+ 705  continue
+!
+! Loop over nuclei - Nuclear attraction part ---
+      do 703 n=1,natom
```
manuelF: Where avoidable, please please please try to not use do <line> continue kind of loops and just do proper indentation. 
The flow of the code is much more complicated to follow if you have them.

This is valid for all the file, and might be able to do in a following PR.

lioamber/TD.f
```diff
@@ -92,6 +92,8 @@ SUBROUTINE TD()
 ! Group of charges
        groupcharge=.false.
 !!------------------------------------!!
+	nopt=0
```
manuelF: Why this?

lioamber/input_read.f90
```diff
@@ -55,6 +55,8 @@ subroutine read_options(inputFile, charge)
                    cut2_0, cut3_0,                                             &
                    ! Distance Restrain
                    number_restr,                                               &
+                   ! Dbug
```
manuelF: Same

lioamber/write_output.f90
Nick-Chaos: It was a bug fix. In lioamber/properties.f90 we have:

call write_forces(dxyzqm, natom-1, 1, uid)dxyz(natom+1, 3)

but dimention of dxyz is (natom, 3) so we need to use  dxyz(natom+1, 3)

ramirezfranciscof: Sorry to intervene, I was just checking the subroutine and I think it would be much better fix to have the loop run from offset to offset+atom-1 rather than giving natom-1 as input, which locally is renamed natom again, and allocating natom+1.

lioamber/int1_new.f90
Nick-Chaos: this is a copy of int1.f subroutine that dont use RMM array. im going to modify identation at the end of process

lioamber/int1_new.f90
Nick-Chaos: same. this is a copy of  int1.f with a few modifications

lioamber/int1_new.f90
Nick-Chaos: im not going to rewrite all subroutine, just modify it to not use RMM

lioamber/int1_new.f90
Nick-Chaos: same. it is part of int1.f original code

lioamber/TD.f
Nick-Chaos: It is a bug in TD.f variable is used but not asigned

lioamber/init_lio.f90
Nick-Chaos: modified

lioamber/write_output.f90
Nick-Chaos: I agree, but i didnt chage all subroutine because i dont know why modified to use an offset

ramirezfranciscof: Ok, an alternative solution would be to have the offset start in 0 instead of 1 and then have the loop go from 1+offset to natom+offset. I think this would be the most natural way of using the offset as an offset.

lioamber/liomods/garcha_mod.f
Nick-Chaos: Done

lioamber/write_output.f90
Nick-Chaos: done on b3891f1e49891a43150544dd8d9aeb3fdbb71b84

Nick-Chaos:
Fixed errors in subroutine obtain_new_P and re-tested with gfortran.

lioamber/write_output.f90
fedepedron: I believe it will, I just wrote this subroutine by copy-pasting from the old "main.f90". The offset is needed since it prints twice: once from 1 to natom, the second from natom, natom+nsol. I believe I wrote it that way for... reasons...

ramirezfranciscof: Generalizing both writes by using an offset is a really nice abstraction, but the loop has to start with a +1 (and thus I'm guessing that your comment should read "the second from natom+1, natom+nsol"). Too much C can make you think counting starts at 0 but that's for crazy people. Also, the meme is "**because of** reasons". Ta- ta.

fedepedron: I know my memes, but strictly speaking for can be used when looking at the consequence and not the causes. Nevertheless I changed it hoping that it somehow makes you happy (?). Coming from C/Java/Ruby, I still have trouble when starting from 1 in FORTRAN. Guess I'm crazy people.

ramirezfranciscof:
Have you tried running dynamics with amber? I am getting the following error:

At line 1030 of file SCF.f90
Fortran runtime error: Attempting to allocate already allocated variable 'fockm'

I compiled with just "make"; I will explore it further after lunch, but if you see this @Nick-Chaos , try to check if you can detect if there's any obvious problem with the code modifs.

fedepedron:
Regarding FFR's comment, the following variables are allocated but never properly deallocated:  
Y, Ytrans, Xtrans, scratch, scratch1 (these are deallocated only if DIIS,  but always allocated)
Dvec, Vmat, Xmat, Ymat, Xtrp, Ytrp
atom_group, orb_group, orb_selection
fockm, FP_PFm, bcoef
znano, fockbias
Whether this is the cause of the error, I'm not sure; but they should be anyways properly deallocated. If they are used for further calculations, the deallocation should be performed in liomain().

lioamber/write_output.f90
```diff
@@ -64,10 +64,9 @@ end subroutine write_dipole
 subroutine write_forces(dxyz, natom, offset, uid)
     implicit none
     integer, intent(in) :: uid, natom, offset
-    real*8 , intent(in) :: dxyz(natom, 3)
+    real*8 , intent(in) :: dxyz(natom+offset, 3)
```
manuelF: I am not sure the set of changes in here and properties.f90 should be in this PR. So far the rest of the PR is more separating and cleaning some names, whereas this is a functional change. 
Should these two changes be part of this PR?

Nick-Chaos: The original change ( natom +1) was necesary to compile with -traceback -check all -fp-stack-check flags and test code. Other changes were made before the comments here. Should we not perform these changes in this PR?

Sorry if this is no the correct place to do a question, but why can I only reply Manuel's comments?

lioamber/SCF.f90
```diff
@@ -552,232 +904,151 @@ subroutine SCF(E)
...
+!
+! here in RMM(M5) it is stored the new Fock matrix
+! test damping on Fock matrix
+!
+!      DAMP=gold
         if (niter.eq.1) then
           DAMP=0.0D0
+	  hagodiis=.false.
+	  if (DIIS ) then ! agregado para cambio de damping a diis, Nick
+            allocate(fockm(MM,ndiis),FP_PFm(MM,ndiis),bcoef(ndiis+1))
```
ramirezfranciscof: This (line 1030) is where I'm getting my error. Why was this allocation introduced? I'm getting complaints that fockm is already allocated.

Nick-Chaos: i fixed it and im testing it at this moment, need a few minutes for push new commit.
The problem was the SAVE Statement of some arrays.

lioamber/write_output.f90
Nick-Chaos: fixed it on c4310b3900f0993e95d0c8e717f563225f0099d7. and tested with amber

PR #78, opened 2017-02-06
Dipole subroutine cleanup and update

fedepedron:
Updated the dipole subroutine and improved readability. Please check if you approve this PR:

- [x] Nano

- [x] Fran
- [ ] Nico
- [ ] Manu
- [ ] Uri


fedepedron:
I do not know if this can be fixed now that the damage is done. For future updates, I (we) should use "git mv" insead of "git add/rm" (which I used out of ignorance).

ramirezfranciscof:
I'm not sure that alone could solve the problem. The command `git mv` seems to be [just a shorthand](http://stackoverflow.com/a/1094392/638366) for the `mv`/`git add`/`git rm` series of commands. Further [proof](http://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv/3010099#3010099). If it wasn't detected this way, probably wouldn't have been detected with `git mv`. I still think this has to be solved in some other way, like making commits or even pull requests with just the renaming (as per the link in my previous post).

Nick-Chaos:
t have compilation problems with: make intel=1 cublas=1 time=1
need to remove dipmem.o in make_depend.mk

PR #79, opened 2017-02-07
Bug fix: wrong dimension in nnat and 2 variables had not initialized …

Nick-Chaos:


fedepedron:
May be there is a better way to do this than hardcoding nnat size, but well.

lioamber/init_lio.f90
```diff
@@ -169,7 +169,7 @@ subroutine init_lio_common(natomin, Izin, nclatom, charge, callfrom)
              cx(ngdDyn,nl) , ax(ngdDyn,nl)     , Nucx(ngdDyn), ncontx(ngdDyn) ,&
              cd(ngdDyn,nl) , ad(ngdDyn,nl)     , Nucd(ngdDyn), ncontd(ngdDyn) ,&
              indexii(ngDyn), indexiid(ngdDyn)  , v(ntatom,3) , Em(ntatom)     ,&
-             Rm(ntatom)    , af(ngdDyn)        , nnat(ngDyn), B(ngdDyn,3))
+             Rm(ntatom)    , af(ngdDyn)        , nnat(200)   , B(ngdDyn,3))
```
manuelF: Why 200? Where does this constant come from?

test/12_ECP/Cl2/restart.out
```diff
@@ -1,32 +1,32 @@
...
--0.6671445E+00   0.1089919E+00   0.7250910E+00
- 0.1560502E-07  -0.1650665E-07  -0.8036178E-07  -0.6671447E+00
--0.1172172E+00  -0.7250909E+00   0.1089920E+00
- 0.5066089E-01  -0.5437683E-03  -0.5544642E+00   0.6476632E-07
- 0.4985568E-07  -0.2505318E-06   0.2725456E-06
--0.1223954E-08  -0.1222385E-08  -0.7218613E-07   0.1193533E+00
--0.6793012E+00   0.1108266E+00   0.7372970E+00
- 0.2042327E-07  -0.1365305E-07  -0.6887784E-07  -0.6793012E+00
--0.1193532E+00  -0.7372967E+00   0.1108267E+00
+-0.6854849E+00  -0.7330578E+00  -0.1452930E+00  -0.8304452E-08
```
manuelF: Why this change? Shouldn't this change make all the restart.out for every test change as well?

ramirezfranciscof:
Just as a comment, we accepted fairly quickly this last pull request because we wanted to check what happened when a modification was added while there was still a pull request in queue (fede's).

nanolebrero:
Usually the atom of the same type (e.g. all the oxigen atoms) have the same
basis functions types (same exponents and coefficients). If all the atoms
type share their functions the number of elements in the periodic table
should be enough. 200 give as the ability to use differents functions types
for different atoms of the same type.

2017-02-07 17:16 GMT-03:00 Manuel Ferreria <notifications@github.com>:

lioamber/init_lio.f90
Nick-Chaos: nnat dimention is the number of kinds of atoms in the basis file. (121 max if include most rare and unused elements, an over dimention may allow to include extra basis for same atom).
The original dimention ngDyn had problems  in small systems because ngDyn was the number of contractions for the system and this may be smaller than kinds of atoms in basis file.
Dimention of nnat may be dynamic again in future, we may add that as an issue.

test/12_ECP/Cl2/restart.out
Nick-Chaos: that file is not used as test, actually we may remove it

PR #82, opened 2017-02-15
Changes to properties file.

fedepedron:
Made a minor relocation of the do_xxxx (xxxx being dipole, forces, fukui and population) since it will be really useful for unit tests. That said, such routines are meant to be only temporary until we achieve better ways to organize stuff.

### DIFF
--- a/lioamber/liomain.f90
+++ b/lioamber/liomain.f90
@@ -149,7 +149,7 @@ subroutine do_population_analysis()
    ! Performs Löwdin Population Analysis if required.
    if (lowdin) then
        do kk=1,natom
-           q(kk)=real(Iz(kk))
+           q(kk) = real(IzUsed(kk))
        enddo
        call lowdin_calc(M,natom,RealRho,sqsm,Nuc,q)
        call write_population(85,natom,IzUsed,q,1)
--- a/lioamber/liomain.f90
+++ b/lioamber/liomain.f90
@@ -170,6 +170,8 @@ subroutine do_fukui()
     real*8  :: fukuim(natom), fukuin(natom), fukuip(natom), softness
 
     if (OPEN) then
+        ! TO-DO
+        ! Add call fukui_calc_OS when openshell is available.
     else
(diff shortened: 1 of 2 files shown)


lioamber/liomain.f90
```diff
@@ -49,3 +54,131 @@ subroutine liomain(E, dipxyz)
...
+!%% DO_FUKUI %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+! Performs Fukui function calls and printing.                                  !
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+subroutine do_fukui()
+    use garcha_mod, only : X, NCO, M, natom, Nuc, Smat, Eorbs, Iz, OPEN
+
+    implicit none
+    real*8  :: fukuim(natom), fukuin(natom), fukuip(natom), softness
+
+    if (OPEN) then
```
manuelF: If you don't have a 'then' branch, then just negate the condition. Otherwise leave a TODO here to add the Open branch,

fedepedron: Added the to-do comment

fedepedron:
The thing is, those routines added to liomain will be reduced in the future, since most of the code can be greatly simplified (they are just calls for calculations and printing routines).  However, I could not keep it in properties.f90 since it was hell for unit tests.

ramirezfranciscof:
I think the most certain organizational advantage is that now properties only has calculations (instead of calculations + managerial interfaces) so it is cleaner. Including them on liomain is currently inocuous, since liomain is actually empty and unused (the second best thing would be to delete it), but might also be a second possible convenience in the future: since the yet unfulfilled purpose of liomain was to be a general managerial interface, the content of these added subroutines might eventually be passed onto liomain.

PR #83, opened 2017-02-16
Unit Tests - Part I

fedepedron:
Fixed existing unit tests and wrote new ones for most of the properties (except for Fukui which needs careful planning).


ramirezfranciscof:
@manuelF Github seems to not be very smart in recognizing movement of files. Commit 41ed38310c28a43bcdb79ea174a5637ffeb066b0 seems to recognize that the file was moved but in the pull request the relationship between the deleted file and the new one is lost. Is there a way to prevent this?

fedepedron:
Just a little comment, it is not needed to recompile all of LIO to test this pull request (since nothing in the main program has changed). In fact, you only need to run the runtests.sh script in the utests folder.

ramirezfranciscof:
Perhaps we could check out one of [github's integrated tools](https://github.com/integrations/feature/code-quality) that looks at code coverage.

lioamber/utests/test-ftermbias.f90
```diff
@@ -0,0 +1,94 @@
...
+    ! Tests for null SQS matrix.
+    sqsmat = 0.0d0
+    vector = 1
+    weight = 1.0d0
+    outmat = 0.0d0
+    testResult = "FAILED"
+    
+    call fterm_biaspot(3, sqsmat, vector, weight, outmat)
+    testCond = (maxval(abs(outmat)) < criteria)
+    if (testCond) testResult = "PASSED"
```
manuelF: What if it doesn't pass? How do you know?

lioamber/properties.f90
```diff
@@ -87,7 +87,6 @@ subroutine mulliken_calc(N, M, RealRho, Smat, NofM, q0, q)
     integer :: i, j, k
     real*8  :: qe
 
-    call g2g_timer_start('mulliken')
```
manuelF: These seem unrelated changes, revert?

fedepedron: It is needed since the test are intended to run independently. If those lines are not deleted, libg2g is required.

lioamber/utests/test-ftermbias.f90
fedepedron: I'm not sure if I understand the question, but by default testResult is "FAILED". If it doesn't pass, then it won't be changed to "PASSED" and will print "FAILED".

lioamber/properties.f90
manuelF: Can you then not install a fake libg2g with these two functions for the tests only? Just like it is done for the other modules.

fedepedron: I think it's not worth the effort since the time will be removed from that routine anyways... However, I was about to change mulliken_calc in an upcoming PR, so we can keep this now (and the test will be momentaneusly broken) and then delete the lines.

manuelF: I was saying this because it will happen eventually with other routines that call g2g functions like this. We should fake them out if they are not relevant, so tests are kept minimal.

lioamber/utests/Makefile
```diff
@@ -1,28 +1,17 @@
 ######################################################################
 # MAKEFILE FOR TESTS
 ######################################################################
-FC := gfortran
 
-tests :=
-tests += vvfterm_test01.x
-tests += lowdinpop_test01.x
+FC = gfortran
```
Nick-Chaos: May we include an option for choose compiler. I mean i would like to do test with same compiler that i use for lio.

lioamber/utests/test-properties.f90
```diff
@@ -0,0 +1,275 @@
...
+    nDeg       = 0
+    nDegMO     = 0
+    testResult = "FAILED"
+
+    call get_degeneration(energies, nOrb, M, nDeg, nDegMO)   
+    testCond = (nDeg.eq.1).and.(nDegMO(1).eq.1)
+    if (testCond) testResult = "PASSED"
+    write(*,*) testResult, ' - Usage of a single orbital.'
+    deallocate(energies, nDegMO)
+
```
Nick-Chaos: This test fails, nDegMO(1)=0. It is ok?
If we expect obtain nDegMO(1)= 0 should not be PASSED?
or maybe have to change special case for M=1 in get_degeneration

lioamber/utests/Makefile
fedepedron: Yes; I intended to do this in the following update, if you want I'll add it right away

lioamber/utests/test-properties.f90
fedepedron: It is intended, thanks to this I realized a minor bug in get_degeneration subroutine lol. Upcoming fix for this.

PR #84, opened 2017-02-20
Minor bugfixes and improvements to properties calculation.

fedepedron:
Changed mulliken_calc so it only needs one variable for charges, and fixed the routine calls accordingly. Also minor related changes to other routines in properties and liomain.

### DIFF
--- a/lioamber/properties.f90
+++ b/lioamber/properties.f90
@@ -77,7 +77,7 @@ end subroutine get_degeneration
 subroutine mulliken_calc(N, M, RealRho, Smat, NofM, q)
 
     ! RealRho         : Rho written in atomic orbitals.                !
-    ! q0, q           : Starting and final Mulliken charges.           !
+    ! q               : Starting and final Mulliken charges.           !
     ! M, N, NofM, Smat: N° of basis functions, atoms, Nuclei belonging !
     !                   to each function M, overlap matrix.            !
     implicit none
--- a/lioamber/properties.f90
+++ b/lioamber/properties.f90
@@ -29,6 +29,7 @@ subroutine get_degeneration(energ, orb, M, nDeg, nDegMO)
          
     if (M.eq.1) then
         nDeg = 1
+        nDegMO(1) = 1
         return
     endif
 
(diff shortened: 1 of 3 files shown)


lioamber/properties.f90
```diff
@@ -73,25 +74,20 @@ end subroutine get_degeneration
 !%% MULLIKEN_CALC %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 ! Performs a Mulliken Population Analysis and outputs atomic charges.          !
 !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
-subroutine mulliken_calc(N, M, RealRho, Smat, NofM, q0, q)
+subroutine mulliken_calc(N, M, RealRho, Smat, NofM, q)
 
     ! RealRho         : Rho written in atomic orbitals.                !
     ! q0, q           : Starting and final Mulliken charges.           !
```
manuelF: Remove q0  and update the comment now that it no longer exists

lioamber/properties.f90
```diff
@@ -29,6 +29,7 @@ subroutine get_degeneration(energ, orb, M, nDeg, nDegMO)
          
     if (M.eq.1) then
         nDeg = 1
+        nDegMO(1) = 1
```
manuelF: Why this change?

fedepedron: That is because the get_degeneration routine returns nDeg (degeneration number) and nDegMO, which are the orbitals degenerated. If M=1, then nDeg=1 but also nDegMO should be set to 1.

manuelF: Change looks unrelated to the rest of the PR (which is just functional changes) and are non-obvious. Either make a separate PR or add comments in this if block for what it is doing.

fedepedron: I believe it is related, since it is indeed a bugfix to a property calculation. If this is not implemented, nDegMO(1) will return 0 which is not correct.

PR #85, opened 2017-02-24
Exodus to F90 - Chapter 1

fedepedron:
I've been modifying some files in order to start the migration to F90. I also added documentation wherever I could. There are no changes in functionality, and the files edited are those which do not use any kind of module (in an upcoming PR, I'll edit those which use modules but not RMM so that Nick doesn't panic).

I have some questions to share though:
Should matmuldiag and read_groups be kept as separate files? In addition, read_groups is not called anywhere, maybe I should just bring up an issue.
Also, in density.f90, last lines, shouldn't there be an "if" branch using garcha_mod: OPEN?


ramirezfranciscof:
What is the reasoning behind commit 5992409 ? (adding subs name at the end) Is there a reason why this would be good practice? Because it would seem to me like unnecessary code duplication.

fedepedron:
There is a reason indeed. On a side note, I believe it improves readability, particularly in files containing several subs; that is not however the reason. If I recall correctly, properly ending subroutines with the name allows debug compilation flags such as checkbounds to perform additional checks. Isn't this correct @Nick-Chaos ?

PR #87, opened 2017-03-14
Exodus to F90 - Chapter 2

fedepedron:
New issue of the "Exodus to F90" series, which will go on hiatus until further developments are incorporated to rc.


lioamber/readECP.f90
```diff
@@ -1,231 +1,227 @@
...
+  write(*,4166)
+  write(*,4167) , ppotat_acum
+  write(*,4168)
+  write(*,*)
+
+4160 FORMAT(4x,"╔═════════════════════════════════════&
       ══════════════╗")
- 4161 FORMAT(4x,"║    READING EFFECTIVE CORE POTENTIAL PARAMETERS    ║")
- 4162 FORMAT(4x,"╠══════════════╦═════════════════════╦&
+4161 FORMAT(4x,"║    readinG EFFECTIVE CORE POTENTIAL PARAMETERS    ║")
```
fedepedron: clearly an issue with find and replace... I'll fix that soon.

lioamber/readECP.f90
```diff
@@ -1,231 +1,227 @@
...
+4162 FORMAT(4x,"╠══════════════╦═════════════════════╦&
       ══════════════╣")
- 4163 FORMAT(4x,"║   ATOM TYPE  ║   NUMBER IN SYSTEM  ║   ECP TYPE   ║")
- 4164 FORMAT(4x,"╠══════════════╬═════════════════════╬&
+4163 FORMAT(4x,"║   ATOM TYPE  ║   NUMBER in SYSTEM  ║   ECP TYPE   ║")
```
fedepedron: same, find and replace

fedepedron:
I believe keeping capital letters in reserved keywords is not an option. First and most important of all, we agreed to remove unnecessary capitalization in the meeting we had last year; and second, I disagree on the matter that improves readability, IMHO it's the other way about. Even more, most of the code is kept in lowercase, while capital letters are sparsely used; in most cases there are even "clusters" of uppercase in greatly lowercase code sectors, which definitely messes up the reading. 

Regarding capitalization of names and variables, the thing I said before about removing unnecessary uppercase stands still. And, come on, is not that hard to use "grep -i". In fact I end up using that all the time since some variables (xnano for example IIRC) are either capitalized or undercased randomly.

Last but not least, I decided to keep variables with Spanish names _for now_. That should also change in the near future, lest it adds even more confusion to an already messy code.

Nick-Chaos:
mmm i don`t remember agreed that, actually i remember that manuel`s code for transform .f to .f90 may auto capitalize code. 

the problem is not in grep case, search in vim (or similar) is easier if same variable is wrote in the same way. The case of xnano and similars needs to be corrected, but change more variables in specific parts of the code only makes the problem bigger.

fedepedron:
Don't get me wrong, I agree that variables (specially module variables), parameters and subroutines should be written consistently throughout the code. But being LIO an open source code with the chance to grow (eventually) into a community, we should abide to current coding standards (and I believe, for example, that upcasing keywords was deprecated at least two decades ago). Correct me if I'm wrong, but currently uppercase words are reserved for parameters/constants and library calls, while the rest is kept in lowercase with underscore/camelCase. In fact, I recall we also decided on using underscores everywhere (even though I wanted camelCase in variables). 

That said, when the time is ripe I will bring upon myself the tedious task of properly casing/underscoring everything, so that all parts of the code are consistent. Of course Nano has the last word on this, but we should start setting aside our particular preferences; we may not be coding LIO forever, so we should keep everything as clean, standardized and consistent as possible for newcomers, so that they may not experience the days of RMM's burning hell ever again.

Nick-Chaos:
Me parece bien que nano decida en que queda esta boludez. Sin embargo sea cual sea la decicion final creo que el cambio de formato hay que hacerlo luego de terminar y mergear todas las versiones de lio como habiamos dicho en un principio ya que seguir cambiando RC hace que sea mas dificil unir las diferentes ramas.

manuelF:
There is an additional issue with randomly changing capitalization/formatting for aesthetic reasons. Git Blame (the tools that tells you which commit changed which line) breaks on reformatting. If we need to change code to fit new standards, then this is something we have to consider. 

Therefore, I think, we should not rename until we do it in a single commit (so it is easy to do git blame before that). If we have multiple, intertwined commits with formatting back and forth, then for sure, this incredibly useful tool becomes useless.

ramirezfranciscof:
> Therefore, I think, we should not rename until we do it in a single commit (so it is easy to do git blame before that). If we have multiple, intertwined commits with formatting back and forth, then for sure, this incredibly useful tool becomes useless.

I agree with this. It makes me a little suspicious of some of the previous pull requests which also dealt with reformatting where I was not paying attention to this. Perhaps we should create a common branch for the reformatting so we can work collectively in this, but still keep it separated from RC until all changes are ready.

Now, there is also another side of this issue. Thou I agree that given the need to preserve the blame as far as possible we should abstain from progressive reformatting the existing code just to reformat it, I also believe that this implies that any new code we add or code we modify for other reasons should be formatted to conform to the established standards from the get-go. Which also means we need to decide and set those standards.

@Nick-Chaos I know you were of the opinion that this could wait, which made sense at the time we were discussing because we had forgotten about this price to pay for the reformatting (and still makes sense for code we are not modifying for other reasons). But I think we all can agree now that the sooner we apply this to the new modifications, the less information we will lose in the process of adaptation.

fedepedron:
I agree with most of the comments, in particular those regarding the use of blame. It still bugs me the fact that we will lose blame sooner or later, and in that case reformatting _before_ doing further changes should be the way to go. 

That said, I believe the most peaceful solution would be what Fran said: keep a separate branch for reformatting and then eventually merge it into rc. In fact, the files I chose to reformat are not vital and most likely won't be changed in the near future, even after including things such as open shell. And it's not just a reformat; in fact what I intended to do was to remove implicits, properly call modules, and add necessary documentation wherever I could; it just seemed a good choice to reformat on the go.

ramirezfranciscof:
@fedepedron : send the request to the newformat branch.

Issue #75, opened 2017-01-27
Compilation flags cleanup


ramirezfranciscof:
We talked about simplifying and reorginizing the compilations flag and discussed some particulars ( issue #66 ) but I think we didn't agreed on an overall scheme. I propose the following ideas:

* **cuda:** 0 doesn't use gpu; 1 uses cuda subs; 2 uses cuda and cublas. If this seems like one option too many, perhaps we could have only 0 or 1 (1 activating all possible gpu tools), which would require that the user has cublas installed to use any gpu optimization. Since these are free libs (as far as I know) perhaps it would not be so restrictive to require them for gpu usage.

* **parallel:** 0 doesn't use parallelization; 1 uses openMP: 2 uses MPI (to be added "soon"); 3 uses both.

* **debug:** 0 means no debug and maximum optimization (-O3, vectorization, etc); 1 means no optimization at all; 2 means use debug flags; 3 use debug flags + internal code options we might add. Perhaps we could do without 2 and have 3 directly.

* **intel**: 0 means use GNU compilers+libs; 1 means use intel compiler; 2 means intel+MKL. I am assuming the MKL make a big difference in performance. If this stops being the case, perhaps we could ditch them.

* **precision:** 0 means use mixed precision, 1 means use full double.

* **verbose:** initially, as explained in issue #66 . However, I am wondering if this doesn't overlap a little with "debug", specially when talking about the values of 5 and above. Perhaps to reconcile this the idea would be that debug tells you what to "calculate" and verbose what to "print out", but it still not very convincing because a dependency still exists (you can't print out what you haven't calculated). Also, do the verbose levels of 5 and above imply the previous ones? (that is to say, if level is 5 it will also print user data that comes with level 3). Below I present an idea to deal with this.

<br>

## Deprecated flags:

* **dbg** is now debug and is more complete.
* **cpu** redundant with cuda?
* **profile** included in debug.
* **full_double** is now precision
* **cpu_recompute:** I am not sure what is the idea of this...
* **time:** not sure what to do with this one, but having one flag just to enable/disable timers seems kind of a waste. I think ultimately it's fate will depend on how we resolve the "verbose" flag. See below.

<br>

## Verbose conflict:

Perhaps we don't need a verbose flag at compilation. Perhaps user information should be requested using input keywords, and programming information should be requested by using "debug" for tools that negatively affect performance (to check functionality) and a new "profile" flag that activates resource light tools that don't affect performance (such as timers) to test for efficiency.

<br>

[...]

PR #91, opened 2017-04-04
Newmake

ramirezfranciscof:
Changed the makefile procedure:

* Changed a bit of the variables to simplify.

* Change in name of options as discussed in issue #75 .

* All flags are now together under said scheme of options (instead of having compilation and linking separated).

* Makefile.optchck is a script-like makefile that checks the option of last compilation for consistency. Also stores the information in module "build_info" so that it can be seen from inside liolib.

Please check all the compilations options you normally use to see if any is broken. When this pull is incorporated, you can check with me to rebase your working branches.


ramirezfranciscof:
I'm gonna change "Makefile.depends" back to "Makefile.depend" just for ease of comparison; before squashing commits please let me know so I can set it back.

lioamber/Makefile.depend
```diff
@@ -80,81 +164,31 @@ tmplist += int1.o int2.o int3lu.o int3mem.o intfld.o intsol.o
...
+tmplist += maskrmm.o write_output.o SCF_exter.o
+$(tmplist:%.o=$(OBJPATH)/%.o) : $(OBJPATH)/garcha_mod.mod
 
 
-# maskrmm: Description pending
-######################################################################
-objects   += maskrmm.o
-src_paths += maskrmm
+################################################################################
+# TO-DO: Erradicate RMM from the face of earth.
```
fedepedron: This (Y)

lioamber/Makefile.depend
```diff
@@ -80,81 +164,31 @@ tmplist += int1.o int2.o int3lu.o int3mem.o intfld.o intsol.o
 tmplist += int1G.o int2G.o int3G.o intSG.o intsolG.o intsolGs.o
 tmplist += jarz.o lio_finalize.o predictor.o
```
fedepedron: Need to add "intECP.o" to tmplist in order to properly compile.

Nick-Chaos:
no me estaría compilando con: make intel=1 time=1 cublas=1
el error esta en los include.
intECP.f90(127): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [GARCHA_MOD]

lioamber/Makefile.depend
fedepedron: @Nick-Chaos check this out: add tmplist += intECP.o to keep testing. @ramirezfranciscof will fix it soon <sup>TM</sup>.

lioamber/Makefile
```diff
@@ -1,131 +1,105 @@
...
-	$(FC) $(FFLAGS) -c $< -o $@
+%.mod : %.o             $(MAKEFILES) | $(OBJPATH)
+	@touch $@
 
-$(obj_path)/%.o   : %.f90 $(makefiles) |  $(obj_path)
+$(OBJPATH)/%.o :  %.f   $(MAKEFILES) | $(OBJPATH)
 	$(FC) $(FFLAGS) -c $< -o $@
 
-$(obj_path)/%.o   : %.F90 $(makefiles) |  $(obj_path)
+$(OBJPATH)/%.o :  %.F   $(MAKEFILES) | $(OBJPATH)
```
manuelF: Can you just drop one of the capitalizations (.F and .F90) and rename (in a different PR, preserving history) to lowercase? It just looks unnecessarily complicated having both supported for every file.

lioamber/Makefile
```diff
@@ -1,131 +1,105 @@
...
-
+$(OBJPATH)/%.o :  %.F90 $(MAKEFILES) | $(OBJPATH)
+	$(FC) $(FFLAGS) -c $< -o $@
 
+liblio-g2g.so : $(OBJECTS:%.o=$(OBJPATH)/%.o)
+	$(LC) $^ $(LFLAGS) $(LIBS) -o $@
 
-$(obj_path) :
-	mkdir -p $@
+#.PHONY: force clean install uninstall
```
manuelF: Either have all the phony in a single line or remove this comment,

lioamber/Makefile.options
```diff
@@ -0,0 +1,171 @@
...
+
+else ifeq ($(cuda),2)
+  DEFINE += -DCUBLAS
+  LFLAGS += -L$(CUDA_HOME)/lib64/
+  LIBS   += -lcudart -lcublas
+  cublas  = 1 # This is necessary for current status of Makefile.depends but
+              # should be changed in the future.
+
+else ifeq ($(cuda),3)
+  DEFINE += -DCUBLAS
```
manuelF: These three flags are defined for all cuda instances. Make them common when cuda is enabled, and not specific to each level of $(cuda)

lioamber/Makefile.options
```diff
@@ -0,0 +1,171 @@
...
+else
+  $(error Invalid value in option intel = $(intel))
+endif
+
+
+################################################################################
+ifeq ($(parallel),0)
+else
+  $(warning This option is not available yet, set it to 0 or avoid it.)
+  $(error Invalid value in option intel = $(intel))
```
manuelF: Should say parallel.

lioamber/Makefile.options
```diff
@@ -0,0 +1,171 @@
...
+  LFLAGS += $(DEBUG01)
+
+else ifeq ($(analytics),-2)
+  #TODO: DPRINT_MATRICES unnecessary, code should recognize DEBUGGING
+  DEFINE += -DDEBUGGING=2 -DPRINT_MATRICES
+  FFLAGS += $(DEBUG02)
+  LFLAGS += $(DEBUG02)
+
+else
+  $(error Invalid value in option intel = $(intel))
```
manuelF: Should say analytics

lioamber/Makefile.options
```diff
@@ -0,0 +1,171 @@
...
+
+################################################################################
+ifeq ($(analytics),0)
+
+else ifeq ($(analytics),1)
+  DEFINE += -DPROFILING=1
+  FFLAGS += -pg
+  LFLAGS += -pg
+
+else ifeq ($(analytics),-1)
```
manuelF: Scary that a flag can be negative. Where is the documentation for this?

lioamber/build_info.f90
```diff
@@ -0,0 +1,115 @@
...
+         '  with different options to have a more reliable run.'
+         write( unit=outunit, fmt= '(A)') ''
+#     endif
+#     ifdef INCOMPLETE_BDATA
+         write( unit=outunit, fmt= '(A)') &
+         '  WARNING! INCOMPLETE DATA:'
+         write( unit=outunit, fmt= '(A)') &
+         '  There has been a problem with the data passed during'
+         write( unit=outunit, fmt= '(A)') &
+         '  compilation. This is a bug and needs to be reported.'
```
manuelF: To whom? Leave instructions.

lioamber/Makefile
ramirezfranciscof: I think traditionally lowercase/uppercase in fortran code extensions was used to imply whether the pre-processor had to be run (.F) or not (.f) ([sauce](http://stackoverflow.com/questions/19743517/fortran-difference-between-f-and-f-extension)). We force pre-processor through compilation options anyways and, honestly, I can't think of a reason someone would not want to pre-process a source code. I added it for compatibility and to consider other people's preferences (or mine? I don't remember which came first, lower or upper case), but I would agree that since we are unifying criteria, this should be standardized.

One thing though: I think it would be better to change it in the same PR as the extensions, in order not to have broken commits. It should not affect the recognition of the file renaming, I think that only happens when you modify and rename the same file.

lioamber/Makefile.options
ramirezfranciscof: Not exactly, it is only enabled for cuda=2 or 3, not cuda=1. However I would agree that it would be preferable if I could do something along the lines of "if cuda > 2, then enable this...", but I found that there is no such a comparison statement in Make. I could define another variable that says "cublas_on" and have is set to true (or one) in the cases cuda=2 and cuda=3, but I think it would not be much different: I would still be duplicating code, it would be less duplication, but I would also be adding more complexity in the form of the extra variable "cublas_on". Can you think of a better way to implement this?

(Note: I know there is already a "cubla_on"-like variable named "cublas" that is also used in Makefile.depends, but I'm actually going to try to eliminate the need for this variable in the future because I want to minimize the "communication" variables that are used in more than one Makefile.xxx).

lioamber/Makefile.options
ramirezfranciscof: Is it a bad idea to have flags be negative? This was an attempt of unifying a "profile" and "debug" flags so as to have less options: positive values of analytics is for increasing the profiling information whereas negative values are for increasing the debugging information. It worked well because you will (almost?) never want to use both kinds of information at the same time. Part of the rationale/discussion can be found in issue #75 , but I was never 100% convinced with this implementation myself. Do you think it would be better to keep the options for profiling and debugging separated?

manuelF: Understood. 

I want to point out here you should really document possible values for every flag: the fact is, there are almost no cases of flags that can be negative in common gnu utils, that's is why it looks so odd. You don't have to do it in the PR, but, eventually, there should be a description for every value. 

Ideally it should, eventually, look like this I think: http://www.fortran.uk/plusfortmanual/cvranal.html#S44

lioamber/Makefile
manuelF: Thanks for the explanation, I didn't know that about fortran. Please then leave both rules here and remove them in a separate PR with the extension renaming, as to keep the purpose of this PR clear.

lioamber/Makefile.options
ramirezfranciscof: Any idea? I think I will be accepting the PR soon and this is the only "issue" still open.

PR #92, opened 2017-04-11
Adendum to Newmake (#91)

ramirezfranciscof:
Planned a separated commit to keep cleaner history, but actually didn't find any file with uppercase extensions.

* Changed Makefile.depend to Makefile.depends

* No more compatibility with uppercase extensions (.F and .F90). Every file is preprocessed.


ramirezfranciscof:
Do you know of any file that still uses uppercase extensions (.F or .F90)?? I couldn't find any.

@nanolebrero @Nick-Chaos @fedepedron @segmentat10nfault

PR #94, opened 2017-04-20
Ehrenfest Beta

ramirezfranciscof:
Already squashed in single commit: changes moved from ehrenfest branch
because rebasing was very conflicting. Not ideal but the feature started
previous to the implementation of better practices.

Usage of ehrenfest features requires to make changes in Amber; the file
to replace is included in the commit:

/dat/ambermod/qm2_extern_lio_module.F90

Testing and improving is still necessary, but it seems to be working so I'm
including it in the code so as not to have too big a divergent branch.


Nick-Chaos:
tested liosolo with ifort and double presition. with and without ECP. all fine

fedepedron:
Tested gfort with cuda and gpu, and with Gromacs. All good.

PR #101, opened 2017-06-12
Fixed unit conversions for SCF_extern and a minor bug with write_forces.

fedepedron:
Topic.


Nick-Chaos:
Tested with intel compilers.

PR #106, opened 2017-07-06
Openshell now fixed for single GPU-only calculations.

fedepedron:
Note: this was done in a branch (numbersnumbers) that was used for a quickfix but was never rebased. Thus the same change is commited twice, once before OS and once after. It was not fixed because when squashing it should become invisible.


ramirezfranciscof:
Rebased! @fedepedron please check it worked correctly before overwriting / rebasing your local copy.

fedepedron:
There seems to be some unfixable issues with this PR, upcoming new more consistent one

PR #114, opened 2017-07-17
Transport1

charlyqchm:
Hi, I added modifications in TD.f and SCF.f90 to implement transport calculation in LIO. For this, a new module have been added, lioamber/transport.f90, and a new routnie, lioamber/cublasmath/cumsp_r.f.
Please, accept my pull request.


lioamber/TD.f
```diff
@@ -494,9 +638,18 @@ SUBROUTINE TD()
          call g2g_timer_sum_stop('Coulomb precalc')
       endif
 #ifdef CUBLAS
-            call CUBLAS_FREE(devPtrY)
+            if(.not.TRANSPORT_CALC) call CUBLAS_FREE(devPtrY) ! charly: no se que quiso hacer uriel acá
```
manuelF: This block is deallocating a matrix stored in the GPU (line 594). If you don't deallocate it it will cause a memory leak.

lioamber/TD.f
```diff
@@ -494,9 +638,18 @@ SUBROUTINE TD()
...
 #endif
 
+!Tansport: initializing Driging rate to the propagation------------------------!
+      if(TRANSPORT_CALC) then
+         GammaMagnus=driving_rate
+         GammaVerlet=GammaMagnus*0.1
+         write(*,*) ' Driving Rate =', GammaMagnus
```
manuelF: Debug code. Please remove.

lioamber/TD.f
```diff
@@ -508,21 +661,32 @@ SUBROUTINE TD()
...
               if (propagator.eq.1) then
                  t=(istep-1)*tdstep
               endif
               t=t*0.02419
               write(*,*) 'evolution time (fs)  =', t
 !--------------------------------------!
-              call int3lu(E2)
-              call g2g_solve_groups(0,Ex,0)
-              write(*,*) '! step & energy', istep,E
+            if ((propagator.eq.1).or.
```
manuelF: Comment here for the logic of this change.

lioamber/TD.f
```diff
@@ -561,6 +726,11 @@ SUBROUTINE TD()
 ! Here we obtain the fock matrix in the molecular orbital (MO) basis.
 ! where U matrix with eigenvectors of S , and s is vector with
 ! eigenvalues
+
+!charly: agregué el if copiando lo de uriel
```
manuelF: Unnecessary comment.

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+
+   mapmat=0
+   i=0
+   j=0
+   n=0
+   nn=0
+
+    DO i=1,M
+       DO j=1,M
+          IF((group(nuc(i)).eq.1).and.(group(nuc(j)).eq.1)) mapmat(i,j)=1
```
manuelF: if group(nuc(i)) is either 1, 2 or 3 then this whole loop can be written in a single line:

mapmap(i,j) = group(nuc(i)) + ((group(nuc(j))) - 1) *3

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+module transport
```
manuelF: There are no comments describing what any of this does, the outputs and the inputs of any subrouting. Please think of the next pool soul/phd that is trying to understand this code.

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+   implicit none
+   logical              :: transport_calc=.false., generate_rho0=.false., gate_field=.false.
+   integer              :: save_charge_freq=0, ipop
+   complex*8            :: traza0, traza
```
manuelF: Stick to english for variables as well.

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+   real*8               :: driving_rate=0.001, scratchgamma, GammaMagnus, GammaVerlet, re_traza
```
manuelF: Pick a naming convention and stick to it:

alltogetherlowercase
camelCase   < my first choice 
use_snake_case  < my second choice
CapitalCase

manuelF: 0.001 is a very specific constant: should it be external and project wide?

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+#endif
+
+   call g2g_timer_start('electrostat')
+         
+   allocate(rho_scratch(M,M,2))
+   rho_scratch=0
+         
+   DO i=1,M
+      DO j=1,M
+         IF(mapmat(i,j).eq.0) THEN
```
manuelF: This is a _very very very_ complicated way of doing this.

case 0 and 9 are the same but are on different branches
case 1 and 2 and 4 and 5 are the  same but have different parenthesis so it doesn't look like that
case 3 and 7 and 6 and 8 are identical.

Explain what are you trying to accomplish here.

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+         IF((mapmat(i,j).eq.6).or.(mapmat(i,j).eq.8)) THEN
+            rho_scratch(i,j,1)=(0.50D0*rho1(i,j))
+            rho_scratch(i,j,2)=(0.50D0*rhofirst(i,j))
+         ENDIF
+      ENDDO
+   ENDDO
+
+   GammaIny=Gamma0*0.5D0
+   GammaAbs=GammaIny
+   write(*,*) 'GammaAbs,GammaIny =',GammaAbs,GammaIny
```
manuelF: Looks like debug code.

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+            rho1(i,j)= (GammaAbs*rho_scratch(i,j,1))-(GammaIny*rho_scratch(i,j,2))
+         ENDDO
+      ENDDO
+
+!-------Stop if NaN-----------!
+
+   DO i=1,M
+      DO j=1,M
+         IF(rho1(i,j).ne.rho1(i,j)) THEN
+            stop 'Huston, we have a problem'
```
manuelF: Debug code. 

Be descriptive in your error message: There was a NaN, why could there be? Which calculation could produce it? Which coordinates? Is all the matrix NaN? Do the NaN have a specific position (e.g. 1,1; 2,2; 3,3 are NaN)

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+   complex*16, intent(inout) :: rho1(M,M)
+   complex*16, allocatable   :: rho_scratch (:,:,:)
+#endif
+
+   call g2g_timer_start('electrostat')
+         
+   allocate(rho_scratch(M,M,2))
+   rho_scratch=0
+         
+   DO i=1,M
```
manuelF: Pick either lower case fortran or upper case (suggestion: lowercase like most of LIO)

lioamber/transport.f90
```diff
@@ -0,0 +1,161 @@
...
+         ENDIF
+      ENDDO
+   ENDDO
+
+   DEALLOCATE(rho_scratch)
+
+   call g2g_timer_stop('electrostat')
+
+end subroutine electrostat
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
```
manuelF: All these subroutines look pretty self contained, can you write a test for these? To make sure they do what you intend to do.

Nick-Chaos:
El codigo no compila con intel, falta alguna dependecia o similar:

ifort liosolo.f90 -o liosolo -fopenmp  -fpp  -L../g2g -lg2g -L../lioamber -llio-g2g -module ../lioamber/obj -I../lioamber/obj
../lioamber/liblio-g2g.so: referencia a `cumsp_r_' sin definir

Nick-Chaos:
la prueba de compilacion siempre la hago sobre un codigo recien bajado del repositorio

segmentat10nfault:
entonces le sacaron un ifdef a algo que use cumsp_r...
cumsp_r es una subrutina que usa cublas y que solo se compila con CUBLAS=1.
Este error no tiene nada que ver con intel sino con que falta un ifdef en
algún lado que diga que  si no compilaron con CUBLAS=1 no usen cumsp_r..
Me juego que esta en transport el problema por que ahi estuve muy peresoso
y no le puse ifdef a sabiendas de que siempre iba a correr con CUBLAS....
Perdon por eso!

Nick-Chaos:
pero estoy usando cublas=1

make intel=1 time=1 cublas=1 full_double=1

segmentat10nfault:
entonces están linkeando mal cublasmath en algún lado o no están llamando
al modulo cublasmath... si queres fijarte si el error esta en transport
hace que el makefile no la compile y fíjate si con eso anda

charlyqchm:
Para compilar yo uso time=1 intel=2 cuda=2, no se si el comentario sirve.

2017-07-18 17:15 GMT-03:00 Uriel Morzan <notifications@github.com>:

gonzalodm:
che nico creo q para que compile con cublas hay que poner cuda=2

2017-07-18 19:45 GMT-03:00 charlyqchm <notifications@github.com>:

lioamber/input_read.f90
```diff
@@ -34,6 +34,9 @@ subroutine read_options(inputFile, charge)
                            FOCK_ECP_write, ecp_full_range_int, Fulltimer_ECP,  &
                            cut2_0, cut3_0
 
+
+    use transport ,only  : transport_calc, generate_rho0, gate_field,          &
```
fedepedron: Are we sure this does not break when compiling only cpu? Maybe a preprocessor **if** should be added here. We must be sure that these additions do not break current functionality.

lioamber/lowdinpop.f
```diff
@@ -0,0 +1,29 @@
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
```
fedepedron: This whole file is unnecessary since Löwdin population has been moved to properties.f90. In addition, it is currently broken in rc since we do not get the proper SQS matrix.

PR #120, opened 2017-08-08
Energy fix

Nick-Chaos:
Corrected energy in QMMM calculations and added a test for QM with puntual charges.

tested with intel and gfortran


lioamber/dft_get_mm_forces.f
```diff
@@ -93,6 +93,7 @@ subroutine dft_get_mm_forces(dxyzcl,dxyzqm)
        call g2g_timer_summary()
        call g2g_timer_clear()
 
+
```
fedepedron: Really? Jajajajaja

test/14_QM_in_Pcharges/basis
```diff
@@ -0,0 +1,52 @@
+gaussian
```
fedepedron: Is this file really necessary? Shouldn't be easier just putting int_basis=t? or is it a custom basis?

Nick-Chaos: yes, i just copy 0_aguaGPU test and added point of charges

lioamber/dft_get_mm_forces.f
Nick-Chaos: obiamente el cambio mas importante

PR #123, opened 2017-08-28
included subroutines for hybrid-lio

Nick-Chaos:
Included 2 subroutines for make QMMM calculations using hybrid program.

### DIFF
--- a/lioamber/SCF_exter.f90
+++ b/lioamber/SCF_exter.f90
@@ -175,3 +175,82 @@ end subroutine ehren_in
       return
       end subroutine
 !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+!%% SCF_hyb %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+! Performs SCF & forces calculation calls from hybrid                          !
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+
+subroutine SCF_hyb (hyb_natom, mm_natom, hyb_r, E, fdummy, Iz_cl,do_SCF, do_QM_forces)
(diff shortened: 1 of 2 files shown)


lioamber/init_lio.f90
```diff
@@ -428,3 +428,33 @@ subroutine init_lioamber_ehren(natomin, Izin, nclatom, charge, basis_i         &
...
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+!%% INIT_LIO_HYBRID  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+! Subroutine init_lio_hybrid performs Lio initialization when called from      !
+! Hybrid software package, in order to conduct a hybrid QM/MM calculation.     !
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+subroutine init_lio_hybrid(hyb_natom, mm_natom, charge, iza)
+    implicit none
+    integer, intent(in) :: hyb_natom !number of total atoms
+    integer, intent(in) :: mm_natom  !number of MM atoms
+    integer             :: dummy
```
manuelF: Why do you have a dummy parameter?

lioamber/SCF_exter.f90
```diff
@@ -175,3 +175,82 @@ subroutine SCF_gro(E, qmcoords, clcoords, clcharge, nsolin)
...
+        fdummy(j,i)=fa(j,i)
+      end do
+      do i=1, mm_natom
+        fdummy(j,hyb_natom+i)=fmm(j,i)
+      end do
+    end do
+
+    return
+end  subroutine SCF_hyb
+
```
manuelF: Too many newlines at end.

lioamber/init_lio.f90
Nick-Chaos: it is same case as init_lio_gromacs and init_lio_amber subroutines, all need to call read_options with a dummy parameter

lioamber/SCF_exter.f90
Nick-Chaos: my bad, sorry

Nick-Chaos:
No in lio, all these routines are made for use lio as external library in "hybrid" program, so test and other stuff will be included in a future official version of hybrid.

ramirezfranciscof:
I think we still should include some sort of test for this sections in our own code. Maybe some program that specifically uses liolib by calling these subroutines with specific parameters and expecting an specific output? Kind of a "systemic" unit test. Not saying it has to be now, but eventually all code should be tested.

PR #124, opened 2017-09-11
Openshell - Second PR

ramirezfranciscof:
Creation and incorporation of common procedures for cpu/gpu.


ramirezfranciscof:
Interested parties please test as much/soon as possible. It has been a long time since we did this and it could be useful for other developments, so I'm going to accelerate the incorporation (unless corrections are requested, I'll be merging this PR next Monday)

g2g/Makefile
```diff
@@ -88,11 +96,11 @@ depend:
 	@rm -f Makefile.depends.bak
 
 libg2g.so: $(OBJ)
-	#@if [ -z "$(CUDA_HOME)" ] ; then echo "CUDA_HOME is not set. Please, check it."; exit 1; fi;
-	#@if ! test -f $(CUDA_HOME)/lib/libcudart.so; then echo "libcudart.so can not be found in CUDA_HOME/lib/, current value is '"$(CUDA_HOME)/lib"'. Please check if CUDA_HOME variable is correctly set."; exit 1; fi;
+#@if [ -z "$(CUDA_HOME)" ] ; then echo "CUDA_HOME is not set. Please, check it."; exit 1; fi;
```
manuelF: No change, revert these lines

g2g/Makefile.common
```diff
@@ -1,6 +1,6 @@
 # -*- mode: make -*-
 
-CXXFLAGS+= -fopenmp -fPIC -Wall -Wextra -Wshadow -I. -I$(CUDAHOME)/include
+CXXFLAGS+= -fopenmp -fPIC -Wall -Wextra -Wshadow -Wno-unused-function -I. -I$(CUDAHOME)/include
```
manuelF: I think these compiler flags deserve their own PR, since they can affect more than just this change, no?

g2g/Makefile.cuda
```diff
@@ -170,17 +170,25 @@ endif
 
 ## Unified GPU Object
 UGO          := cuda/lio_gpu.o
+CUDA_OBJ     :=
```
manuelF: Why this variable definition / modifcation split?

g2g/Makefile.pointxc
```diff
@@ -0,0 +1,11 @@
+# EXTERNALS
```
manuelF: Does this file really belong to g2g/ main folder? Why isn't it part of the Makefile.depends itself?

g2g/cpu/iteration.cpp
```diff
@@ -14,7 +14,11 @@
 #include "../partition.h"
 
 #include <stdlib.h>
-#include "cpu/pot.h"
+#include "pot.h"
+#include "../pointxc/calc_ggaCS.h"
+#include "../pointxc/calc_ggaOS.h"
+#include "../pointxc/calc_ldaCS.h"
+//#include "..pointxc/calc_ldaOS.h"
```
manuelF: Commented code, remove.

g2g/cpu/iteration.cpp
```diff
@@ -93,7 +97,9 @@ template<class scalar_type> void PointGroupCPU<scalar_type>::solve_closed
         partial_density += Fi * w;
       }
       scalar_type exc = 0, corr = 0, y2a = 0;
-      cpu_pot(partial_density, exc, corr, y2a, iexch);
+//      cpu_pot( partial_density, exc, corr, y2a, iexch );
```
manuelF: Same

g2g/cpu/iteration.cpp
```diff
@@ -128,7 +134,9 @@ template<class scalar_type> void PointGroupCPU<scalar_type>::solve_closed
         scalar_type ww2xc = 0, ww2yc = 0, ww2zc = 0;
 
         const scalar_type * rm = rmm_input.row(i);
-        #pragma vector always
+        #if INTEL_COMP
```
manuelF: do you need this ifdef? Aren't pragmas not understood by the compiler just ignored?

g2g/cpu/iteration.cpp
```diff
@@ -169,7 +177,9 @@ template<class scalar_type> void PointGroupCPU<scalar_type>::solve_closed
       const vec_type3 dd1(tdd1x,tdd1y,tdd1z);
       const vec_type3 dd2(tdd2x,tdd2y,tdd2z);
 
-      cpu_potg(pd, dxyz, dd1, dd2, exc, corr, y2a, iexch);
+//      cpu_potg(pd, dxyz, dd1, dd2, exc, corr, y2a, iexch);
```
manuelF: Commented code

g2g/cuda/iteration.cu
```diff
@@ -5,6 +5,8 @@
 #include <math_constants.h>
 #include <string>
 #include <vector>
+#include <limits>
```
manuelF: keep these <> imports sorted.

g2g/cuda/iteration.cu
```diff
@@ -25,8 +32,8 @@ texture<float, 2, cudaReadModeElementType> rmm_input_gpu_tex2;
 #endif
 /** KERNELS **/
 #include "gpu_variables.h"
-#include "kernels/pot.h"
-#include "kernels/pot_open.h"
+//#include "kernels/pot.h"
```
manuelF: Commented code

g2g/cuda/kernels/accumulate_point.h
```diff
@@ -1,4 +1,6 @@
 #define WIDTH 4
+//#include "../../pointxc/calc_gga.h"
```
manuelF: Commented code, remove?

g2g/cuda/kernels/accumulate_point.h
```diff
@@ -36,8 +38,10 @@ __global__ void gpu_accumulate_point_open(scalar_type* const energy, scalar_type
          _dd2_b += dd2_b[this_row];
      }
   }
-  gpu_pot_open<scalar_type>(_partial_density_a, _partial_density_b, _dxyz_a, _dxyz_b, _dd1_a, _dd1_b, _dd2_a, _dd2_b,
-      exc_corr, exc, corr, corr1, corr2, v_a, v_b);
+//  gpu_pot_open<scalar_type>(_partial_density_a, _partial_density_b, _dxyz_a, _dxyz_b, _dd1_a, _dd1_b, _dd2_a, _dd2_b,
```
manuelF: Commented code, remove?

g2g/cuda/kernels/accumulate_point.h
```diff
@@ -46,11 +50,11 @@ __global__ void gpu_accumulate_point_open(scalar_type* const energy, scalar_type
     energy_c1[point] = ((_partial_density_a + _partial_density_b) * point_weight) * corr1;
     energy_c2[point] = ((_partial_density_a + _partial_density_b) * point_weight) * corr2;
   } else {
-    energy[point] = 0.0f;
+ /*   energy[point] = 0.0f;
```
manuelF: Commented code, remove?

g2g/cuda/kernels/accumulate_point.h
```diff
@@ -89,7 +93,9 @@ __global__ void gpu_accumulate_point(scalar_type* const energy, scalar_type* con
      }
   }
 
-  gpu_pot<scalar_type, compute_energy, true, lda>(_partial_density, _dxyz, _dd1, _dd2, exc_corr, y2a);
+//  gpu_pot<scalar_type, compute_energy, true, lda>(_partial_density, _dxyz, _dd1, _dd2, exc_corr, y2a);
```
manuelF: Commented code, remove?

g2g/cuda/kernels/energy_open.h
```diff
@@ -91,7 +102,7 @@ __global__ void gpu_compute_density_opened(
     }
 
 
-    for (int bj = min_i2; bj >= 0; bj -= DENSITY_BLOCK_SIZE)
+    for (int bj = 0; bj <= min_i2; bj += DENSITY_BLOCK_SIZE)
```
manuelF: Why do you change the iteration order here? Looks like it had a specific order before.

g2g/fix_compile.h
```diff
@@ -0,0 +1,10 @@
+#if !GPU_KERNELS
+#   define __host__
+#   define __device__
+#else
+#   include <cuda_runtime.h>
+#endif
+
+#if !INTELCOMP
```
manuelF: Does nothing.

g2g/matrix.cpp
```diff
@@ -46,7 +48,8 @@ template<class T> void HostMatrix<T>::alloc_data(void) {
   }
   else
   {
-    posix_memalign((void **) &this->data, 64, this->bytes());
+    posix_return = posix_memalign((void **) &this->data, 64, this->bytes());
+    if (!(posix_return == 0)) { std::cout <<"HostMatrix: Error in posix_memalign.\n"; };
```
manuelF: This doesn't even compile.
Also, indent like the rest of the code.
if (posix_return != 0) {
  std::cout <<"HostMatrix: Error in posix_memalign.\n"; 
  exit(1); <==== missing exit, otherwise what do you do? do you continue when this alloc failed?
}

g2g/partition.cpp
```diff
@@ -403,12 +405,12 @@ void Partition::solve(Timers& timers, bool compute_rmm,bool lda,bool compute_for
...
   #pragma omp parallel for num_threads(cpu_threads+gpu_threads) schedule(static) reduction(+:energy)
   for(uint i = 0; i< work.size(); i++) {
-    bool gpu_thread = false;
 #if GPU_KERNELS
+    bool gpu_thread = false;
     if(i>=cpu_threads) {
       gpu_thread = true;
       cudaSetDevice(i-cpu_threads);
-    }
+    }
```
manuelF: Trailing space.

g2g/pointxc/calc_ggaCS.h
```diff
@@ -0,0 +1,282 @@
...
+#include <math.h>
+#include <float.h>
+#include <cassert>
+
+#include "../fix_compile.h"
+#include "../scalar_vector_types.h"
+
+
+namespace G2G {
+#define POT_ALPHA     ((scalar_type)-0.738558766382022447)
```
manuelF: Weren't all these constants defined elsewhere already?

g2g/pointxc/calc_ggaCS.h
```diff
@@ -0,0 +1,282 @@
+#include "pbeCS.h"
```
manuelF: keep <> imports first, and then " "

g2g/Makefile
fedepedron: It is a change in fact, now these lines do not appear in screen output when compling. Though I must agree is a sort of irrelevant change for this PR

g2g/Makefile.common
fedepedron: Same as before, I agree with this.

g2g/Makefile.cuda
fedepedron: @ramirezfranciscof

g2g/Makefile.pointxc
fedepedron: I believe we intended to keep it separate in order to reduce clutter in the .depends file.

g2g/cpu/iteration.cpp
fedepedron: Yes, but same as before, it prints a couple of lines to output and this makes it cleaner. And same I before, I agree it is not mandatory for this PR

g2g/cuda/kernels/energy_open.h
fedepedron: Yes, we talked with Nano and decided that the way it was performed before had no impact in performance; this way at least it improves readability

g2g/pointxc/calc_ggaCS.h
fedepedron: I believe it's because they are not needed elsewhere, they are pretty specific for the PBE XC funcitonal.

g2g/matrix.cpp
fedepedron: It did compile for me in several computers... I'll fix it that way though, thanks.

g2g/Makefile.cuda
ramirezfranciscof: This can be reverted if you deem it inappropriate for this PR, but:
(1) You will either need to change the order in which cuda/analytic_integrals and pointxc are added or initialize CUDA_OBJ with POINTXC_OBJ
(2) Initially all cuda code seems to have been separated in the cuda folder, but as more specific (analytic_integrals) and versatile (pointxc) code has been and will be added, it is clear the the cuda objects will be spread in modules grouped under other criteria. C/Cpp and cuda seem sufficiently integrated so that this shouldn't become a problem, but in the future this will call for a makefile structure that better reflects this (for example, having an object list for all cuda objects with a clean slate initialization letting you know this is a place to accumulate stuff). Again, not saying this should be done now, just putting it out there.

g2g/pointxc/pbeOS_exch.h
```diff
@@ -0,0 +1,91 @@
...
+
+template<class scalar_type> __host__ __device__
+void pbeOS_exch( scalar_type rho, scalar_type s, scalar_type u, scalar_type v,
+                 scalar_type& ex, scalar_type& vx)
+{
+   // Construct LDA exchange energy density
+   scalar_type rho13  = cbrt( rho );
+   scalar_type exunif = (scalar_type)EASYPBE_AX * rho13;
+
+//---------------------//
```
manuelF: Commented out fortran code here? Please remove.

g2g/pointxc/pbeOS_exch.h
```diff
@@ -0,0 +1,91 @@
...
+   // Construct PBE enhancement factor and calculate Ex
+   scalar_type s2 = s * s;
+   scalar_type s3 = s * s2;
+   scalar_type p0 = (scalar_type)1.0f + EASYPBE_UL * s2;
+
+   scalar_type fxpbe = (scalar_type)1.0f + EASYPBE_UK - EASYPBE_UK/p0;
+   ex = exunif * fxpbe;
+
+//---------------------//
+// WHAT IS THIS???
```
manuelF: Commented out fortran code here? Please remove.

test/0_aguaGPU/forces
```diff
@@ -0,0 +1,3 @@
+    1   -0.033308   -0.063064    0.043860
```
manuelF: Are these forces part of the golden files? They are not input of the tests, does it need to be submitted?

test/0_aguaGPU/salida.ref
```diff
@@ -0,0 +1,570 @@
+<====== Initializing G2G ======>
```
manuelF: This file shouldn't be submitted in this PR. It is just output, and not necessary to run any tests

g2g/pointxc/calc_ldaCS.h
```diff
@@ -0,0 +1,134 @@
...
+      ex = 0; ec = 0;
+      y2a = 0;
+      return;
+   }
+
+   scalar_type y = cbrt( dens );  // rho^(1/3)
+   scalar_type v0 = (scalar_type)-0.984745021842697 * y; // -4/3 * (3/PI)^(1/3) * rho^(1/3)
+   ex = POT_ALPHA * y; // -(3/PI)^(1/3) * rho^(1/3)
+
+   switch(iexch) {
```
manuelF: What are the values of iexch? 1,2,3? Where are these described? Can these be an enum?

g2g/pointxc/calc_ggaOS.h
```diff
@@ -0,0 +1,116 @@
...
+   scalar_type vxpbe_a = 0.0f;
+   scalar_type vxpbe_b = 0.0f;
+   scalar_type ecpbe   = 0.0f;
+   scalar_type vcpbe_a = 0.0f;
+   scalar_type vcpbe_b = 0.0f;
+ 
+   exc_corr = exc = corr = corr1 = corr2 = v_a = v_b = 0.0f;
+
+   scalar_type dgrad_a, delgrad_a, rlap_a, dgrad_b, delgrad_b, rlap_b;
+   if ( (dens_a + dens_b) > ((scalar_type)1e-13f) ) {
```
manuelF: the 1e-13f epsilon, can this be a constant?

g2g/pointxc/calc_ggaOS.h
```diff
@@ -0,0 +1,116 @@
...
+      if (Vxc_id == 9) { //PBE
```
manuelF: Vxc_id is also a constant for which XC method to use, where are all these defined? Can they be constants?

g2g/pointxc/calc_ggaCS.h
```diff
@@ -0,0 +1,282 @@
...
+void calc_ggaCS( scalar_type dens, 
+                 const vec_type<scalar_type,width>& grad,
+                 const vec_type<scalar_type,width>& hess1,
+                 const vec_type<scalar_type,width>& hess2,
+                 scalar_type& ex, 
+                 scalar_type& ec, 
+                 scalar_type& y2a)
+{
+   // hess1: xx, yy, zz  || hess2: xy, xz, yz
+   if (dens < 1e-13) { ex = ec = 0; return; }
```
manuelF: Same comment about epsilon here.

g2g/pointxc/calc_ldaCS.h
fedepedron: iexch could range from 1 to any number, but in most places it's just 9 (9 = use PBE XC functional). Other options only work with CPU/Closed shelll calculations, and I am not really sure about that either.

g2g/pointxc/calc_ggaOS.h
fedepedron: Same as iexch, currently useless since we only use iexch=9 (PBE). We made these routines by adapting the existing potential routines (in cpu/pot and cuda/kernels/pot), and we thought making this kind of changes in this PR would make things even messier. Perhaps a good follow-up PR would be "removing everything that is not PBE".

g2g/pointxc/calc_ggaOS.h
ramirezfranciscof: Yes we can.

ramirezfranciscof: So, I changed it but only slightly: instead of just using the number, I define the "minimum_density_value" just before and use that constant. The thing is that I'm not sure how to further generalize this: sure, this number should be the same for both cases mentioned here, but should be the same for every time we want to check if a number is practically zero? Perhaps reorganizing constants, specially these operative ones (where to define them, where to use each) may be left for a specialized pull request focused on that.

ramirezfranciscof:
> This is OK, but I mean to have a global place to keep these constants. Something like math_constants.h where we define the epsilons of these this, to be consistent on cut-off criteria.

> This can definitely be a separate PR since it might be a bit of work on standarization.

I agree @manuelF , just commented almost the same thing in the previous correction, we probably wrote our comments at the same time.

PR #125, opened 2017-09-19
Openshell - Third PR

ramirezfranciscof:
Currently Working:
* CS calculations in all configurations.
* OS calculations in 1 GPU and all AINT_GPU_LEVELs.


lioamber/SCFop.f
```diff
@@ -388,18 +404,40 @@ subroutine SCFOP(E,dipxyz)
...
+      if (igpu.gt.2) then
+         call aint_coulomb_init()
+      endif
+
+      if (igpu.eq.5) MEMO = .false.
       if (MEMO) then
          call g2g_timer_start('int3mem')
-         call int3mem() 
+         call int3mem()   !*1
+!         call int3mems()  !*2
```
manuelF: Remove?

lioamber/SCFop.f
```diff
@@ -388,18 +404,40 @@ subroutine SCFOP(E,dipxyz)
 #endif
 c
 c End of Starting guess (No MO , AO known)-------------------------------
```
manuelF: Choose either c or ! to comment code

lioamber/SCFop.f
```diff
@@ -1389,8 +1315,10 @@ subroutine SCFOP(E,dipxyz)
      >           EMATa,EMATb,EMAT2_a,EMAT2_b,bcoef_a,bcoef_b,suma)
       endif
 
-      deallocate (kkind,kkinds)
-      deallocate(cool,cools)
+      if(MEMO) then
+        deallocate (kkind,kkinds)
```
manuelF: extra space between deallocate and arguments

fedepedron: SCFop (I believe) will be removed in the near future once we join it with SCF; that is why we changed only what was needed or what really improved readability. The rest was kept as it was or copied from SCF.f90 in order to check which parts were almost the same.

ramirezfranciscof:
Corrected the spacing because it was easy to deal with, but I'm leaving the other two comments because:
1) I think commenting inconsistency should be dealt separately and specifically because it will modify the blame history drastically.
2) Though it is not being used right now, the int3mem/int3mems has to be rethinked specifically later and there are some comments that may be relevant. We should soon be dealing with this, but I would rather let it be for now.

PR #132, opened 2017-10-05
Integrals in separated module

ramirezfranciscof:
* faint_cpu modules created to hold analitic integrals (Fortran Analitic INTegrals).
 * TODO: have them use all parameters and not use external mods.
 * TODO: move them all from faint_cpu77 (old format) to faint_cpu (f90 format).
 * TODO: have them each in a separated module so that they can be compiled independently and then included in a "recopilation" module (so that modifying one of them will not require the rest to be recompiled).


ramirezfranciscof:
ADDENDUM: This PR has a problem with what energies are returned in the first steps. I'm not making an issue because I'm pretty sure the problem has been dealt with, I just want to keep in the record that the problem exists here in case this is used for debugging purposes in the future.

PR #134, opened 2017-10-09
TD works without CUBLAS library

gonzalodm:
en TD.f90 la linea rho1=matmul(x,rho) generaba error ya que la maztriz x y rho no son del mismo tamaño, lo unico que hize es dar las dimensiones correctas a x

### DIFF
--- a/lioamber/TD.f90
+++ b/lioamber/TD.f90
@@ -112,7 +112,7 @@
        ALLOCATE(factorial(NBCH))
 !!------------------------------------!!
 ! Mulliken
-       ipop=1
+       ipop=0
 !!------------------------------------!!
         nopt=0
 
@@ -210,7 +210,6 @@
       IF(TRANSPORT_CALC) THEN
          call mat_map(group,mapmat, Nuc, M, natom)
       ENDIF
-
 !--------------------------------------------------------------------!
       if(propagator.eq.2) allocate (F1a(M,M),F1b(M,M))
 !--------------------------------------------------------------------!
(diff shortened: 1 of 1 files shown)


lioamber/TD.f90
```diff
@@ -112,7 +112,7 @@ SUBROUTINE TD()
        ALLOCATE(factorial(NBCH))
 !!------------------------------------!!
 ! Mulliken
-       ipop=1
+       ipop=0
```
ramirezfranciscof: Is this change necessary? Is it relevant for this PR?

lioamber/TD.f90
```diff
@@ -333,15 +331,14 @@ SUBROUTINE TD()
        dovv=.true.
        if (dovv.eqv..true.) then
         if (.not.allocated(orb_group)) then
-          allocate(orb_group(M))
+          allocate(orb_group(M),group(M)) ! agrege que allocate la variable group
```
ramirezfranciscof: Are you doing this because you detected that `group` was not allocated? That variable should contain information. I am fearing that somehow an unpolished version of the atom potential implementation got inserted here. I really don't like the `dovv=.true.` above; did you try deleting that instead of allocating `group`?

gonzalodm:
Perdon estos cambios no deberian estar aqui. Pero ipop=1 hacia que el
codigo se rompa, ya hable con charly y me dijo que lo iva a ver.
Con respecto a group, efectivamente no estaba allocateada

2017-10-09 18:24 GMT-03:00 ramirezfranciscof <notifications@github.com>:

PR #147, opened 2017-10-13
Atompot PR0 - Reorganization and addition of subroutines

ramirezfranciscof:
New subroutines added and old subroutines adapted. Next step would be using it in the code to replace the existing implementation. This can be squashed.


fedepedron:
Tested CPU-only and hybrid cpu/gpu for OS and CS.

PR #122, opened 2017-08-18
Transport1

charlyqchm:
In this version, transport was added to @gonzalodm TD version and problems with the presence of unnecessary routines were solved.

PR #148, opened 2017-10-13
Dftb pr 0

charlyqchm:
* Modules with subroutines and data for future use in the DFTB feature.
* Modules with subroutines and data for convergence acceleration of the SCF cycle. 
* Important simplification of SCF structure. Inclusion of the convergence acceleration subroutines.

QUESTION: This branch was already being developed before the inclusion of the recent Transport1 PR, so it was built upon a series of commits that was later squashed. Upon rebasing, whenever conflicts arose when trying to re-apply these commits, we chose to skip them. But some of those were automatically applied without conflict: you can check that the first 8ish commits in this PR already appeared in the Transport1 PR #122 and were automatically re-applied during the rebase.

@manuelF do you know why this is? is there a better way to do the rebasing forcing it to skip the first N commits automatically?


fedepedron:
This does not compile, throws the following error:

make[1]: Circular obj/lowdinpop.mod <- obj/lowdinpop.o dependency dropped.
m2c    -o obj/lowdinpop.o obj/lowdinpop.mod
make[1]: m2c: Command not found
<builtin>: recipe for target 'obj/lowdinpop.o' failed
make[1]: *** [obj/lowdinpop.o] Error 127
rm obj/lowdinpop.mod

fedepedron:
Tested cpu-only and cpu/gpu. Please rebase before merging.

PR #151, opened 2017-10-17
Ehrenfest PR02 - Incorporation of multiple time-step scheme

ramirezfranciscof:
As the name indicates. I would this to be incorporated with a merge commit rather than squashing.

### DIFF
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -678,7 +678,7 @@
       if (doing_ehrenfest) then
          call spunpack('L',M,RMM(M1),RealRho)
          call fixrho(M,RealRho)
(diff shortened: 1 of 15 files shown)


fedepedron:
Tested with cpu-only and cpu/gpu, everything in order.

ramirezfranciscof:
@gonzalodm @charlyqchm do your thing.

gonzalodm:
Tested with intel compiler and cublas library TD and ehrenfest dynamics.
All fine!!

charlyqchm:
Pull request tested with Cublas and Intel options. SCF, TD and Transport are working correctly

PR #155, opened 2017-10-30
It's timers' time!

fedepedron:
Timers are now usable as a runtime option, rendering the "time" compilation flag obsolete. The current scheme adds the "timers" integer variable to lio namelist, which can get the following values:

timers = 0  => No timers at all.
timers = 1  => Only single start/pause/stop timers are used.
timers = 2  => Only timer summary is used.
timers = 3  => Both single timers and timer summary are used.

Please test at will.

### DIFF
--- a/g2g/Makefile
+++ b/g2g/Makefile
@@ -46,14 +46,6 @@ ifeq ($(print), 1)
         CXXFLAGS += -DPRINT_MATRICES
 endif
 
-ifeq ($(time), 1)
-  CXXFLAGS += -DTIMINGS
-endif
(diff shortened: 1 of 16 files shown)


charlyqchm:
checked pull request with Cublas and Intel options. SCF, TD and Transport are still working XD

gonzalodm:
I´ve tested this PR with gfortran and intel with cublas and without cublas
and. It´s works perfectly

2017-11-06 12:39 GMT-03:00 charlyqchm <notifications@github.com>:

PR #159, opened 2017-11-18
Continous integration with Travis CI

manuelF:
This PR creates a basic continous integration using www.travis-ci.org

What this does is create a basic compile test (make -j cuda=0 cpu=1) that is run on every commit and pull request to LIO. This is a sanity check that the code compiles and has no major warnings using existing flags.

This is just a baseline so then we can eventually add cuda=1 and other compilation flags. Ideally, this would also run unit tests and integration tests (like running 0_aguaCPU) so we have some certainty that the code does what we think it does, and we have fewer regressions.


fedepedron:
No performance problems, tested in many configurations, mainly cpu/gpu without intel.

manuelF:
Thanks! I'll merge it then.

PR #161, opened 2017-11-23
Changing TD dipole moment

fedepedron:
Really small PR, changed the way dipole moment is written in TD. It should be compatible with TDanalyze3 now, @Nick-Chaos please have a look at this.


lioamber/write_output.f90
```diff
@@ -65,6 +65,32 @@ subroutine write_dipole(dipxyz, u, uid, header)
...
+    integer, intent(in) :: uid
+    logical, intent(in) :: header
+
+
+    open(unit = uid, file = "dipole_moment_td")
+    if (header) then
+       write(UID,*)
+       write(UID,*) '# TIME (fs), DIPX, DIPY, DIPZ (Debye)'
+       write(UID,*)
+    else
```
manuelF: Why else? It should print the header *and* the variable contents.

fedepedron: The dipole printing is continuous in TD. If you check TD.f90, you will see that the routines is called first only to prepare the file and write the header. Then, each TD step writes the time and dipole moment under certain conditions, so it should not be printing the header every time the subroutine is called.

ramirezfranciscof:
Right now I'm running in eneas so I can't do it, but next week I want to check if I can change my dipole printer in ehrenfest for the one used here and unify things and add it to this PR.

Issue #158, opened 2017-11-16
Makefile dependencies of g2g


gonzalodm:
The pointxc dependencies are not set correctly. Should be:

```
cpu/iteration.o: calc_ggaCS.h calc_ggaOS.h calc_ldaCS.h
cuda/iteration.o: calc_ggaCS.h calc_ggaOS.h calc_ldaCS.h
```

Also, accumulate_point is not a cpp but a header, so there is no .o. It is not clear how to set that chain of dependencies when headers include other headers.

ramirezfranciscof:
[This article](http://www.cplusplus.com/forum/articles/10627/) may be relevant for how to better organize header files. I'm still not sure how to solve the header dependency on other headers problem...

ramirezfranciscof:
Once #162 is merged and this is closed, we should open a new issue to decide how to deal with headers.

Issue #113, opened 2017-07-14
Makefile Bugs


ramirezfranciscof:
To accumulate bugs in the makefile:
- [x] cublasmath should depend on general_module and it does not.
- [x] Include the path to cuda stubs in the makefile (see issue #53).
- [x] Link libg2g and liblioamber to the corresponding path without LD_LIB_PATH
- [x] Add  -Wl,--no-undefined and link all libs.

(see below for clarifications)

ramirezfranciscof:
Include the path to cuda stubs in the makefile.

ramirezfranciscof:
Link libg2g and liblioamber to the corresponding path without LD_LIB_PATH. See [this](https://stackoverflow.com/questions/2726993/how-to-specify-preference-of-library-path) and [this](https://stackoverflow.com/questions/3840218/how-to-specify-the-library-version-to-use-at-link-time) or [this](https://stackoverflow.com/questions/8835108/how-to-specify-non-default-shared-library-path-in-gcc-linux-getting-error-whil) or [this](https://stackoverflow.com/questions/828053/how-do-you-link-to-a-specific-version-of-a-shared-library-in-gcc)

ramirezfranciscof:
Changes:

-LFLAGS += -fPIC -shared
+LFLAGS += -fPIC -shared -Wl,--no-undefined

-  LIBS   += -llapack
+  LIBS   += -lblas -llapack

-  LIBS   += -lmkl_lapack95_lp64 -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core
+  LIBS   += -lmkl_blas95_lp64 -lmkl_lapack95_lp64
+  LIBS   += -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core

ramirezfranciscof:
Clean commits with all warnings

ramirezfranciscof:
I'm using this issue to dump all makefile "bugs" so that I can fix them all in one commit (It has happened a few time I made a correction only to find out the day after that there was another problem that I could have dealt with). I can fix all these once the OS-PR series is complete.

Each post is a different modification, which one do you want me to explain? I think the only one that is confusing is the last one, in which I mean that when activating all warnings (analytics=2 if not mistaken) the compilation should be as clean as possible (yes, this should require a major restructure of many parts of the code; probably should be a whole separate issue or even many PRs).

fedepedron:
Good, in that case, shouldn't we include #112 here?

[...]

PR #162, opened 2017-11-24
Bugfix

ramirezfranciscof:
Dealt with problems of issue #113 and #158


fedepedron:
Tested with several compilation options, although not with Intel.

ramirezfranciscof:
@charlyqchm @gonzalodm this PR should fix the long compilation time of the faint module. It's in your best interest to test this so we can merge.

charlyqchm:
I've tested the pull request with Intel and Cublas. There is no problem with SCF, TD and Transport, but I can't note difference between the new compilation time and the last one.

PR #172, opened 2018-01-25
Fixing TD issues and reformating.

fedepedron:
I know this PR is big and includes a huge change in TD, so please DO test it as hard as possible.
@charlyqchm please check if transport runs propperly, since I can only do unit tests at most.
In an upcoming PR I will properly encapsulate TD with its own module variables and subroutines, also editing predictor and magnus subs.

Change log:
- Reformatted TD dismembering it in several subroutines and deleting unnecessary code.
- Encapsulated most of transport related subroutines.
- Fixed TD not working in CPU.
- Fixed TD not working with Ifort.
- Fixed TD dipole output issues with ifort.


lioamber/transport.f90
```diff
@@ -18,35 +19,46 @@ module transport
...
+   integer, intent(out) :: ngroup, ntdstep
+   real*8 , allocatable, intent(inout) :: overlap(:,:)
    integer, allocatable, intent(inout) :: group(:), mapmat(:,:)
-   integer :: orb_group(M)
+   integer :: orb_group(M), icount
    real*8  :: GammaMagnus, GammaVerlet
 
-   ngroup = 0
+   ngroup  = 0
+   ntdstep = 1
```
charlyqchm: What is supposed to do this? For this reason Transport is not working. Without that, SCF, TD and Transport are working without problem. I tested with Cuda=2 and Intel=2.

P.S.: How do you bug this? jaja

charlyqchm:
The information in the file "MullikenGroup" is not beeing printed correctly.

fedepedron:
How should it be printed then?

lioamber/transport.f90
fedepedron: Copy pasting, most likely LOL

PR #175, opened 2018-01-30
Overlap PR3 - Bulk of SCF cleanup

ramirezfranciscof:
* Reorganized the content of obtain_new_P so that it is no longer a necessary structure.
* Variable cleanup inside SCF: xnano, znano, some rmmXX, etc.
* Procedure cleanup inside SCF: externalized some details of the implementation of ECP and cubegen; reorganized together dovv (soon to be fockbias); eliminated dipole moment final calculation.


ramirezfranciscof:
DUDE! look at the title you use when you make the merges. I'll see if I can ammend it.

PR #177, opened 2018-01-31
TD and Transport encapsulation

fedepedron:
With some luck this is the last PR regarding TD and transport organization, save perhaps for the upcoming propagator and field modules.

@charlyqchm  do your trick with the transported man. @gonzalodm  was facing problems with cublas=2 intel=1 compilations for small molecules, but that involves the SCF PRs @ramirezfranciscof  has been doing lately.

### DIFF
--- a/lioamber/Makefile.depends
+++ b/lioamber/Makefile.depends
@@ -7,7 +7,7 @@
 # modules (or rather their corresponding '.o' files).
 #
 # When adding new content, you should:
-# 
+#
 #  (1) Include in the list OBJECTS the object file to be included.
 #
 #  (2) Specify the dependencies related to this object, preferably adding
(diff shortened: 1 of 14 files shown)


ramirezfranciscof:
By changing the merge commit it is no longer recognized by your branch, claiming to have conflicts. I'm not sure that rebasing will skip the repeated commit or if it will apply it by doing nothing. I would recommend to (rebase inteactively)[https://stackoverflow.com/questions/12604824/how-do-i-rebase-while-skipping-a-particular-commit] (`git rebase -i rc`). In any case, you will need to force the pull on your local rc.

fedepedron:
Seems that by just rebasing it fixed itself \o/

charlyqchm:
I compiled with Cuda=2 and Intel=2. Everything seems to be ok. SCF, TD and Transport are working normally. But you didn't correct the problem of printing MullikenGroup during the 200 first steps.

PR #178, opened 2018-02-01
Overlap PR3.5 - Cleanups by hiding CUBLAS

ramirezfranciscof:
First steps towards hiding cublas switch properly: added mask_cublas
module and introduced subroutines there that take care of initialization
if cublas is enabled or do nothing if it is not. This is a first step,
not the final result: there are still some ifdefs in SCF that we need
to eliminate.

Left as a separated PR to make it stand out from the rest of the cleanup.

### DIFF
--- a/lioamber/Makefile.depends
+++ b/lioamber/Makefile.depends
@@ -204,8 +204,7 @@ include liosubs_dens/liosubs_dens.mk
 SRCDIRS += liosubs_dens
 OBJECTS += liosubs_dens.o
 #
-#modusrs :=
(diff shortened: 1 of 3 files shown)


ramirezfranciscof:
Weren't we using the D&D version progression?

ramirezfranciscof:
Serious Mode: it is 3.X because it could perfectly be just part of the SCF restructuring that I did with the previous PR, but I wanted to signal it out because of how it dealt with modularizing a compilation option.

PR #180, opened 2018-02-02
Overlap PR4 - Fixed all overlap decompositions

ramirezfranciscof:
Overlap should work with:
* Cholesky decomposition (default: `lowdin = .false.`).
* Lowdin or Symetric orthogonalization (`lowdin = .true.`).
* Canonical orthogonalization (right now it can only be enabled through code modifications).

We should set a better input for this in a future PR.


ramirezfranciscof:
For future reference: the content of this PR was incorporated as PR #181.

PR #182, opened 2018-02-05
Fockbias PR0 - Bias potential on SCF

ramirezfranciscof:
New implementation of fockbias (ex-atompot) is ready for a non-spin dependent potential.
(if open shell, only call fock apply twice for each matrix).

TODO: Actually add the potential to open SCF, TD and Ehrenfest.

TODO: Should be easy to make a spin-dependent potential (different charge
for the different fock matrices). The question if it would be useful...
(¿maybe for a "magnetic field bias" instead of an "electric field bias"?)


lioamber/fockbias_subs.f90
```diff
@@ -0,0 +1,291 @@
...
+      module procedure fockbias_apply_c
+      module procedure fockbias_apply_z
+   end interface fockbias_apply
+
+   contains
+!
+!
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+subroutine fockbias_setup0( activate, do_shape, grow_start, fall_start, setamp )
+
```
fedepedron: IIRC, we said we should avoid needless variable "copy-pasting". Is this routine really necessary?

lioamber/fockbias_subs.f90
```diff
@@ -0,0 +1,291 @@
...
+         print*; stop
+      end if
+   end if
+
+end subroutine fockbias_loads
+!
+!
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+subroutine fockbias_apply_d( timepos, fockmat )
+   use fockbias_data, only: fockbias_is_active, fockbias_is_shaped &
```
fedepedron: This thing of " & <enter> &," instead of ",&" is driving me crazy haha.

lioamber/fockbias_subs.f90
ramirezfranciscof: What do you mean "variable copy-pasting"? Not sure I understand.

I added this initializator in case we have lio used exclusively as a library. IE: instead of relying on a lio.in input, if we want an external program to decide and set this parameters. This is kind of an accessor in OO terms.

lioamber/fockbias_subs.f90
ramirezfranciscof: I think I had a reason to do that. I'm going to check in my fortran books to see if it was there...

lioamber/fockbias_subs.f90
fedepedron: Ok, it is intended to be used externally then, that makes sense.

liosolo/liosolo.f90
```diff
@@ -44,6 +45,21 @@ program liosolo
...
+       print*, "No lio.in file to read."
+
+    else if ( mystat == 3 ) then
+       print*, "No lionml namelist in file lio.in"
+
+    else if ( mystat /= 0 ) then
+       print*, "A problem occurred while trying to read namelist lionml."
+       print*, "stat = ", mystat
+       stop
+
```
fedepedron: I cannot run a single test because everything stops with stat=4.

ramirezfranciscof: Awesome. Right on it.

fedepedron:
Everything seems to be working, but you should rebase.

PR #191, opened 2018-03-09
Operators time

charlyqchm:
Added operator type
This object includes:
- Two matrices representing AO and ON basis, of what is store here.
- Extraction and Storing subroutines for the corresponding matrix.
- Diagonalization subroutine.
- Density building subroutine.
- Conmutation subroutine.
- Base change subroutine from AO to ON.

The most important changes were made in SCF and converger_sub.
This is the first step to include Open Shell in SCF.


fedepedron:
The combination of capital letters/camelCase/underscore in some methods is truly killing me; that aside, everything seems to work fine.

ramirezfranciscof:
Rebasing was straightforward.

PR #195, opened 2018-03-13
Explicitation of variables used in faint subroutines

ramirezfranciscof:
I removed all implicit variables from the faint subroutines and added the "only" clause to the use of garcha_mod. Eventually most of this variables used from garcha_mod should be introduced and returned explicitly as parameters of the subroutine.

Also, introducing the dependencies needed for this I realized the module/pseudo-submodule structure adopted for this part of the code is not convenient. I will probably "revert" to having all of them as stand alone separated modules.

(Note that most of the 1500 lines deleted are from intsolGs)


fedepedron:
With intel compiler, does not compile with ANY option. The most common error I found is:

liotemp/funct.f90(52): error #6377: FUNCTION must be present on the end-function-stmt of an internal or module function
      END
^

ramirezfranciscof:
@fedepedron plz aprove now

PR #198, opened 2018-03-15
Ehrenfest PR1.1 - First attempt to include magnus

ramirezfranciscof:
I have several kind-of-old ehrenfest commits that would need to be included, because though the way there was rocky, I need the last instance of this to use as a base for further work. Some of these middle steps may have some ugly implementations or be non-conforming with more recent established standards. I can make some modifications if you see something that is too flagrant, but otherwise I would ask to let the small things slide. Once this 3 PR's are done*, I'm going to restructure the whole section better.

(* I'm estimating 1.1 - 1.3, it could be one or two more if the changes are too many)


fedepedron:
Everything seems to work fine!

PR #204, opened 2018-03-22
Ehrenfest PR1.5 - Last of old ehrenfest modifications

ramirezfranciscof:
This one is the last of the OLD ehrenfest inclusions, but ehrenfest is far from production ready. From now on I will try to work more compartmentalized and integrate more frequently so I don't have to do these big long many-step inclusions.

Since this is the last PR, I can add modifications freely, so you can unleash all your critiques @fedepedron


fedepedron:
> this is the last PR, I can add modifications freely, so you can unleash all your critiques @fedepedron

Allow me to cite the machinegun philosopher Tychus Findlay: "Hell, it's about time!"

lioamber/basis_subs.f90
```diff
@@ -0,0 +1,154 @@
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
```
fedepedron: Why separating basis_subs from basis_data?

lioamber/ehrensubs/RMMcalc3_FockMao.f90
```diff
@@ -54,17 +55,22 @@ subroutine RMMcalc3_FockMao( DensMao, ElecField, FockMao, DipMom, Energy )
          Qc = Qc + Iz(kk)
       end do
 
-     call intfld( g, ElecField(1), ElecField(2), ElecField(3) )
+      call intfld( g, ElecField(1), ElecField(2), ElecField(3) )
```
fedepedron: Maybe you should consider using the field module for this...

fedepedron:
I have tested this and everything seems to run fine. Even the hexaaquo Fe(II) runs well :stuck_out_tongue:

lioamber/ehrensubs/RMMcalc3_FockMao.f90
ramirezfranciscof: Ok, I'll check that out. Either case, I'm probably going to export the RMMcalc subs back to the module where they should be (maskrmm) on a next step. See #205 .

lioamber/basis_subs.f90
ramirezfranciscof: I have become quite phobic to variables being accessed without being explicitly requested in the smallest scope. Smallest scope in this case being the subroutines contained in the module that can see all variables declared in the module. I want the subroutine to explicitly express which variables are being used through a "use module only" statement.

I could have both modules in the same file though, we talked about this. I'm still not fully convinced by a single file compiling into two different objects (or headers, which are more similar to the .mod files). Mmm...

PR #206, opened 2018-03-28
Initial guess and guessing bugs.

fedepedron:
Added an alternative initial guess option using a modified aufbau principle (both luckily and sadly, it usually performs better than the one we have already...). This works currently for atoms smaller than or equal to Xe, since these are the atoms for which we already have basis sets.

More importantly, fixed quite a few core bugs in g2g which tended to appear when densities approach zero (in fact, this was the reason why NO exploded); I believe this is the most relevant part of this PR.

@ramirezfranciscof , @charlyqchm, @gonzalodm please test this ASAP. I am particularly interested to see how this performs with @charlyqchm's long chains of Ag/H.


g2g/cuda/kernels/rmm.h
```diff
@@ -40,28 +40,21 @@ __global__ void gpu_update_rmm(const scalar_type* __restrict__ factors,
     first_fj = blockIdx.y * blockDim.y;
   }
 
-  bool valid_thread =
-      (i < m && j < m &&
-       i <= j);  // quiero triangulo inferior solamente TODO: sacar esto
-
+  bool valid_thread = (i < m && j < m && i <= j);
+  // quiero triangulo inferior solamente
```
ramirezfranciscof: (1) Maybe add "()" to each distinct condition. I.E.: `( (i < m) && (j < m) && (i <= j) )`
(2) Comments in spanish? I'm very disappoint...

g2g/pointxc/pbeOS_main.h
```diff
@@ -127,23 +131,16 @@ __host__ __device__ void pbeOS_main(
...
   scalar_type g2 = (scalar_type)1.0f - zet;
   scalar_type g2_23 = cbrt(g2 * g2);
   scalar_type g = (g1_23 + g2_23) / (scalar_type)2.0f;
 
-  scalar_type rho13 = cbrt(rho);
-  scalar_type fk1 = cbrt(EASYPBE_PI32);
-  scalar_type fk = fk1 * rho13;
+  rho13 = cbrt(rho);
+  fk1 = cbrt(EASYPBE_PI32);
+  fk = fk1 * rho13;
```
ramirezfranciscof: Shouldn't these be defined during declaration?

charlyqchm:
I tested this pull request. Transport, DFTB and Open Shell seem to be working correctly.

PR #210, opened 2018-04-10
Fixes

fedepedron:
Really small PR concerning two things:
* Recovered translation part in SCF.
* Made limits for XC integrals dependent on the precision used.

### DIFF
--- a/lioamber/faint_cpu/subm_int1.f90
+++ b/lioamber/faint_cpu/subm_int1.f90
@@ -36,11 +36,11 @@
         integer,                       intent(inout) :: natom
 
         double precision, allocatable, intent(in) :: d(:,:)
-        double precision, allocatable, intent(in) :: r(:,:)
+        double precision, intent(in) :: r(natom,3)
         double precision, allocatable, intent(in) :: a(:,:)
         double precision, allocatable, intent(in) :: c(:,:)
-        integer, allocatable, intent(in) :: Nuc(:)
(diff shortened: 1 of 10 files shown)


fedepedron:
@charlyqchm, @gonzalodm please test.

lioamber/faint_cpu/subm_int1.f90
```diff
@@ -36,11 +36,11 @@ subroutine int1(En, RMM, Smat, Nuc, a, c, d, r, Iz, ncont, NORM,&
         integer,                       intent(inout) :: natom
 
         double precision, allocatable, intent(in) :: d(:,:)
-        double precision, allocatable, intent(in) :: r(:,:)
+        double precision, intent(in) :: r(natom,3)
```
ramirezfranciscof: Are you getting any error because of this? This should not be necessary...

ramirezfranciscof: Wowo, I just see this: the problem should not be the assumed shape, but the allocatable descriptor. In fact, unless its being used somewhere below to allocate/deallocate, none of these variables should be described as allocatable...

fedepedron: | none of these variables should be described as allocatable
I know, after this I'll make a new PR with quite a lot of changes to int1.

charlyqchm:
It seems that everything is working normally. I tested transport, TDFTB, and open shell and seem to be running normally, with and without CUDA.

Issue #66, opened 2016-12-22
Implementing verbose levels and output.


fedepedron:
Main output is written in two files: *.log (errors/warnings/timers) *.out (chemical information).
Verbose should be an integer variable. =0 prints only errors/warnings. Values 1-4 are reserved for user-available data, while values 5 and upper include develop/debug information.

Taken from issue #63: *.out file should not be overwritten, but appended to write new information.
Taken from issue #59: Lio density restart name different from default should automatically set VCINP=T.

ramirezfranciscof:
For further research: when compiling with gfortran, write(6,*) and write(*,*) will output to the same unit, but when compiling with intel write(6,*) outputs to the .out of amber.

Case found: SCF 'converged at n iterations' message will be written in different places according to compiler used.

fedepedron:
If I recall correctly, some units have specific meanings in gfortran, we should avoid using them if possible.

EDIT:
"Standard FORTRAN reserves two UNIT numbers for I/O to user. They are:
     UNIT = 5   for INPUT from the keyboard with the READ statement
     UNIT = 6   for OUTPUT to the screen with the WRITE statement "

source: [](http://www.oc.nps.edu/~bird/oc3030_online/fortran/io/io.html)

ramirezfranciscof:
Ok, interesting, I didn't know about the 101-102, we should have it in mind. However, I don't know how to reconcile what you say in edit1 with what is actually happening when you compile with intel: write(6,*) is effectively outputting in a file. Unless reserve doesn't mean "you won't be allowed to use them" but "you should not use them", in which case it would seem that the amber people are the ones kind of breaking the rules here...

fedepedron:
I arrived to the same conclusion: Intel compiler seems to keep units even between different libraries, while gfortran does not. However, there is still a workaround: If we somehow provide LIO with the AMBER output filename, then it's just a matter of opening that file from LIO as an append.

ramirezfranciscof:
But why would we want to write in amber's output? I think it would be better that we write our own files separated from amber's. Actually, what I find most annoying is fact that they seem to have redirected unit 6 to a file, which manifests with intel's compiling behaviour.

fedepedron:
No problem then, I thought you wanted to write to AMBER's log file; in fact I also believe that outputs should be kept separate.

fedepedron:
Reviving this issue due to my tiredness of seeing gigantic nohup files. I'll soon start working on this, so I propose the following verbose levels:

[...]

PR #215, opened 2018-04-20
New Output

fedepedron:
New way to write outputs with verbose levels, in accordance with #66. Please test.
Still missing:
* Proper namelist printing.
* Transport, Ehrenfest and more obscure sections.
@Nick-Chaos way of printing things is still available via "style" boolean keyword.


ramirezfranciscof:
I will look at this with more detail when I get everything installed here, but I just wanted to point out that there is a module (build_info.f90) that contains the information of the last build. You may want to check it out to include that info in the output.

lioamber/SCF.f90
```diff
@@ -67,6 +67,9 @@ subroutine SCF(E)
       use cublasmath , only: cumxp_r
 #  endif
    use initial_guess_subs, only: get_initial_guess
+   use fileio       , only: write_energies, write_energy_convergence, &
+                            write_final_convergence
+   use fileio_data  , only: verbose
```
ramirezfranciscof: I would always try to have the `module_data` be used only by the `module`, and let other subs use only stuff from `module`. For example here in SCF: I would make a sub WRITE_SCF_START and just call that directly instead of checking for verbose level. I would check for the verbose level inside that sub and just return if its 1 or lower.

lioamber/steep.f90
```diff
@@ -161,7 +162,7 @@ subroutine make_E_array(gradient, n_points,Energy, step_size, E, Fmax)
 	  Energy(i)=E
 	end do
 
-	if(verbose) then !ponerle luego un nivel alto de verbose
+	if(verbose .gt. 1) then !ponerle luego un nivel alto de verbose
```
ramirezfranciscof: Same comment as before (use fileio functions instead of fileio_data data structures) + for comparisons use mathematical symbols (==, >, etc) instead of the abbreviated words (.ge., .le., etc). I think these are easier to understand when browsing through the code, and besides you already used them before (in SCF for example).

lioamber/write_output.f90
```diff
@@ -13,7 +13,7 @@
 ! Sets variables up and calls dipole calculation.                              !
 !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 subroutine write_dipole(dipxyz, u, uid, header)
-    use garcha_mod, only : style
+    use fileio_data, only : style
```
ramirezfranciscof: Maybe all of write_output should be in fileio? Is there a reason why we are keeping them separated?

fedepedron:
Short answer to all comments: I already took into account the usage of module variables, this is a sort of intermediate PR in order not to introduce gigantic changes. Ideally, yes, verbose levels and or style (z.B.) will not be available outside the output printing subs. And property printing will be moved to that module too.

fedepedron:
@charlyqchm, @gonzalodm please check if you agree with this.

gonzalodm:
I've compiled with cuda=2 intel=0 and everythings seems to be works
correctly

2018-05-08 8:37 GMT-03:00 Federico N. Pedron <notifications@github.com>:

PR #217, opened 2018-04-26
Hotfix - Bug in gpu XC calculations.

fedepedron:
I'll be accepting this tomorrow morning; @Nick-Chaos found a bug in Exc calculation for gpu kernels (appeared in #206) which ended up in slightly wrong energies and forces.


ramirezfranciscof:
I think it would be good policy that every time we fix a bug we also need to add some kind of test (either a run test or debug flag activated check) to catch it if it appears again.

fedepedron:
In this case it was visible with any kind of test involving forces, although changes were not overwhelmingly large. Gotta be more careful with this. We should start making unit tests though.

fedepedron:
Fixed bug in Exc calculation for gpu kernels (appeared in #206), which ended up in slightly wrong energies and forces.

PR #218, opened 2018-05-22
Quick Change - Activating Fields for transport calculations.

fedepedron:
Opening this on behalf of @charlyqchm. Accepting ASAP.


ramirezfranciscof:
@charlyqchm care to explain a little bit? I mean, why the change? why was the field if-guarded for the case where you weren't doing transport and now it isn't?

charlyqchm:
@ramirezfranciscof This "if" was in the first TD, because the field was on by default, so if you wanted to use transport the field should be off. Now I want to use electric fields, but in specific times of the dynamic.

PR #219, opened 2018-05-22
Output printing - Namelists

fedepedron:
What @ramirezfranciscof has been longing for lately. @charlyqchm @gonzalodm please test.
Upcoming PR will be the last in this line of output standarization.


ramirezfranciscof:
Great work! I have very few issues with this PR, very satisfied with the modifications on the ehrenfest part. Fix the compilation problem, have the little kids run the tests and merge.

Sidenote: I'm still not a 100% sold on how we output the namelist, it's very...idk, hard maintenance?  Like, I wouldn't like to discover that for X reason it is more convenient to have the "=" sign right next to the name of the keyword with no space between them because then I would have to make like a thousand changes. Let me try to put it this way: it feels like it is a fully unrolled loop. Two copies of it actually (write_dull / write_style).

I feel like there should be a way to do this more generically, but the truth is I don't exactly know how =S. IO-handling is the worst, it's such a shame the default nml writting is so shitty.

charlyqchm:
Everything seems to be working correctly, I checked Transport and DFTB, and there is no problem.

PR #223, opened 2018-05-30
Added comments in some parts of the transport module for clarity

IgnacioJPickering:
This is just a test to check how this works. I added comments in some parts of the transport_data and transport_subs modules (which were almost completely devoid of comments) in an effort to understand them.


ramirezfranciscof:
Dude, you don't use the R-word in a semi-institutional public website.

fedepedron:
This is outright censorship, I demand my freedom of speech.

ramirezfranciscof:
I would agree with Mr. Inadi here though, commenting seems a little excessive, specially the kind of comments where you are only saying you are calling a subroutine, which is kind of self evident.

I do like the general header though, I would format it so that lines don't surpass the 80c mark.

IgnacioJPickering:
You probably like the header because I copypasted it from one of your subroutines =) I comment my own code that way because I tend to be stupid myself (?) but I will keep that in mind for future commits.

IgnacioJPickering:
(uuups, I just realized that I kept working on this branch on unrelated stuff after forgetting everything about the pull request =$ what can I do about that?)

fedepedron:
I was about to tell you to keep PRs as much thematically consistent as possible. The short and easiest way to do this (considering the changes are small) is to keep elsewhere a copy of your modified files, undo the last two commits in this branch and start a new branch from RC copying in the modified files.

lioamber/fockbias_subs.f90
```diff
@@ -46,12 +45,17 @@ subroutine fockbias_setup0( activate, do_shape, grow_start, fall_start, setamp )
    real*8 , intent(in) :: fall_start
    real*8 , intent(in) :: setamp
 
+!   fockbias_is_active = activate
```
ramirezfranciscof: Am I understanding this right? The variable initialization is commented but then later is...exactly the same?

ramirezfranciscof:
> (uuups, I just realized that I kept working on this branch on unrelated stuff after forgetting everything about the pull request =$ what can I do about that?)

Not sure what exactly do you mean. You pushed work to your branch without knowing it would automatically be added to the PR? If that is the case, the solution is rather easy:

(1) Go to the last commit (probably cebe100) and open a new branch from there (`checkout -b work_branch`)
(2) Go back to `fockbias_nacho` and point it back at the desired commit (probably 216df7f ). This would be a `reset --hard 216df7f` if I'm not mistaken, but check with @fedepedron.
(3) Force push your changes in "fockbias_nacho" (`push -f ...`).

As a corollary from what @fedepedron said for PRs, it is usually better to keep branches as thematically consistent as possible too. This tends to be hard because you usually notice you needed something to be different in part X when you where working on part Y, but if you notice this then it is better if you can make the changes quickly in another branch, PR that change, and rebase your other branch once it is accepted. Also, yeah, remember that when a branch is involved in a PR, everything you push to it will update the PR (you can keep working locally in your PC-repository, but that's just inviting catastrophe).

On an unrelated notice: I checked the PR and can't find where you "hardcoded" something for fockbias. What do you mean by that?

Issue #224, opened 2018-06-01
Lowdin charges are broken


IgnacioJPickering:
Just run H2O and asked for lowdin charges with the up to date rc branch and got atomic numbers instead, is the same happening to everyone?. Because I am getting atomic numbers I believe this should be a fairly easy thing to fix. Will look into it.

fedepedron:
Addressed as of #275.

PR #225, opened 2018-06-03
Fixing lowdin charges

IgnacioJPickering:
This is an actual PR! =) I am not yet totally sure of how to rebase so I just made the PR directly, there are not that many changes so it will most likely not be a problem. 
This pull request fixes issue #224:

1) **The lowdin_calc subroutine was not working properly when one asked fot lowdin charges and/or mulliken charges on a SP calculation**. When you asked for both LIO would spit out mulliken charges for both. When you asked only for lowdin charges lio would spit out atomic numbers (dumb LIO). 

The reason for this was twofold: first, lowdin_calc was being fed the output partial charges q after a mulliken calculation had been made. Second, lowdin_calc eats sqsmat (the square root of the overlap matrix), but sqsmat was allocated and then subsequently deallocated in the SCF subroutine. Lowdin_calc was therefore being fed zeroes and did nothing with the charges it was input with. 

2) **The lowdin_calc subroutine inside transport was also not working properly.**

The reason for this was that transport subroutine called lowdin_calc but fed it with the overlap matrix instead of sqsmat, and as a consequence the lowdin charges obtained in a transport calculation were rubbish. 

In order to fix this I first reset the partial charges in between mulliken and lowdin calculations in liomain and I also edited lowdin_calc so that it is fed smat and it calculates sqsmat internally. I preferred to do this instead of feeding it with sqsmat directly because **sqsmat is not really calculated anywhere in the program** except when using it for the fockbias subroutines, and I was worried that smat could change dynamically at some point since as far as I could see it is global state. I am not terribly familiar with the TD and steepest descent subroutines, and I was kind of scared that the overlap matrix could change at runtime there and as a consecuence sqsmat should be updated. By feeding it with smat we are certain that we are always using the most up-to-date info. 

If this is not the case and we can ALWAYS think of the overlap matrix as a global constant then it would be more efficient to calculate sqsmat and leave it as a global constant too, then I can modify that quickly. In any case the current way to do it is probably not so bad since I am only building sqsmat everytime lowdin_calc is called, which is hopefully not that often (it perhaps is called often during transport?).

As a side effect of feeding lowdin_calc smat the utests for lowdin_calc are now not working. I don't really know how to fix this because I don't understand them properly, so I will leave that to somebody else =)

**At least Lowdin charges are working properly after these changes.**

### DIFF
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -411,19 +411,18 @@ subroutine SCF(E)
            RMM(M13+kk-1) = Dvec(kk)
         end do
 
-!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%tmp!
 !
 !
 !  Fockbias setup
         if ( allocated(sqsmat) ) deallocate(sqsmat)
         if ( allocated(tmpmat) ) deallocate(tmpmat)
         allocate( sqsmat(M,M), tmpmat(M,M) )
-        !tmpmat and sqsmat should probably be equal on output
+        !tmpmat and sqsmat should probably be equal on output, 
+        !but for some reason they are not...?
         call overop%Gets_orthog_2m( 2, 0.0d0, tmpmat, sqsmat )
-        write(*,*) "tmpmat is", tmpmat
-        write(*,*) "sqsmat is", sqsmat
         call fockbias_loads( natom, nuc )
-        call fockbias_setmat( sqsmat )
+        call fockbias_setmat( sqsmat  )
         deallocate( sqsmat, tmpmat )
 
 
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -418,9 +418,12 @@ subroutine SCF(E)
         if ( allocated(sqsmat) ) deallocate(sqsmat)
         if ( allocated(tmpmat) ) deallocate(tmpmat)
         allocate( sqsmat(M,M), tmpmat(M,M) )
-        call overop%Gets_orthog_2m( 3, 0.0d0, tmpmat, sqsmat )
+        !tmpmat and sqsmat should probably be equal on output
+        call overop%Gets_orthog_2m( 2, 0.0d0, tmpmat, sqsmat )
+        write(*,*) "tmpmat is", tmpmat
+        write(*,*) "sqsmat is", sqsmat
         call fockbias_loads( natom, nuc )
-        call fockbias_setmat( tmpmat )
+        call fockbias_setmat( sqsmat )
         deallocate( sqsmat, tmpmat )
 
 
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -386,13 +386,7 @@ subroutine SCF(E)
         allocate(Xmat(M_in,M_in), Ymat(M_in,M_in))
 
         call overop%Sets_smat( Smat )
-        if (lowdin) then
-!          TODO: inputs insuficient; there is also the symetric orthog using
-!                3 instead of 2 or 1. Use integer for onbasis_id
-           call overop%Gets_orthog_4m( 2, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
-        else
-           call overop%Gets_orthog_4m( 1, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
-        end if
+        call overop%Gets_orthog_4m( 1, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
 
 ! TODO: replace X,Y,Xtrans,Ytrans with Xmat, Ymat, Xtrp, Ytrp
 !        do ii=1,M
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -386,7 +386,13 @@ subroutine SCF(E)
         allocate(Xmat(M_in,M_in), Ymat(M_in,M_in))
 
         call overop%Sets_smat( Smat )
-        call overop%Gets_orthog_4m( 1, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
+        if (lowdin) then
+!          TODO: inputs insuficient; there is also the symetric orthog using
+!                3 instead of 2 or 1. Use integer for onbasis_id
+           call overop%Gets_orthog_4m( 2, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
+        else
+           call overop%Gets_orthog_4m( 1, 0.0d0, X_min, Y_min, X_min_trans, Y_min_trans)
+        end if
 
 ! TODO: replace X,Y,Xtrans,Ytrans with Xmat, Ymat, Xtrp, Ytrp
 !        do ii=1,M
--- a/lioamber/SCF.f90
+++ b/lioamber/SCF.f90
@@ -411,15 +411,13 @@ subroutine SCF(E)
            RMM(M13+kk-1) = Dvec(kk)
         end do
 
-!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%tmp!
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 !
 !
 !  Fockbias setup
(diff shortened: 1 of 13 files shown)


IgnacioJPickering:
Other small issues fixed: the name of the output file was "mulliken", but the lowdin charges were output to the same file when both were asked. When only lowdin charges were requested they were output to fort.85 (!). I also changed some commented code in ehrensubs that seemed to use lowdin_calc so that if it ever gets uncommented it will be paid attention to, and I changed the names of some internal variables in lowdin_calc and mulliken_calc, since the routines did the same thing but had totally different variable names.

fedepedron:
BEEP-BOOP does not compile. Now seriously, the dependency between properties.f90 and typedef_sop is missing in the makefile, please check that.

IgnacioJPickering:
Thanks, it should now be working. Unfortunately there is an extra issue I have found with this fix which I would prefer to talk face to face. The problem is the following: 

TD routines are using sqsmat and passing sqsmat to transport routines. As far as I could see the ONLY reason they are doing this is in order to print lowdin charges. 

I have been told we are pretty sure that smat shouldn't change in the course of any calculation, so a definite fix would probably have sqsmat calculated in the beginning of SCF and then passed on to the corresponding TD routines, and also passed again to liomain. 

The problem is that passing sqsmat to every subroutine seems terribly misguided in my opinion, as a consequence every subroutine has about 30 arguments. This is happening at the moment with TD routines and I would very much like to fix it and not contribute to the problem.

It seems like the alternatives are passing 30 arguments to every routine or using garcha_mod, and both are kind of crappy. I would love to talk about a better way to do this.

lioamber/SCF.f90
```diff
@@ -411,16 +405,18 @@ subroutine SCF(E)
...
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%tmp!
 !
 !
 !  Fockbias setup
         if ( allocated(sqsmat) ) deallocate(sqsmat)
         if ( allocated(tmpmat) ) deallocate(tmpmat)
         allocate( sqsmat(M,M), tmpmat(M,M) )
-        call overop%Gets_orthog_2m( 3, 0.0d0, tmpmat, sqsmat )
+        !tmpmat and sqsmat should probably be equal on output, 
+        !but for some reason they are not...?
```
ramirezfranciscof: I have just checked this, and although it was me who said they had to be equal I was actually wrong. You can remove this comment.

lioamber/SCF.f90
```diff
@@ -411,16 +405,18 @@ subroutine SCF(E)
            RMM(M13+kk-1) = Dvec(kk)
         end do
 
-!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%tmp!
```
ramirezfranciscof: WTF?

lioamber/properties.f90
```diff
@@ -100,27 +101,34 @@ end subroutine mulliken_calc
 !%% LOWDIN_CALC %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 ! Performs a Löwdin Population Analysis and outputs atomic charges.            !
 !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
-subroutine lowdin_calc(M, N, rhomat, sqsmat, atomorb, atomicq)
- 
+subroutine lowdin_calc(N, M, rhomat, smat, atm_of_bfunc, partial_q)
```
ramirezfranciscof: Hahah, you're a re-namer. I like that.

IgnacioJPickering:
Talked about this with charly a bit today and he thinks that calculating sqsmat inside lowdin will damage performance too much in transport calculations, where lowdin charges need to be calculated almost every step, he recommended we transformed properties into a module and had sqsmat and smat as data global to that module, since those are not supposed to change at runtime we could store a copy of sqsmat there after it is calculated once in SCF and then just have lowdin_calc use it when we need to. As a bonus we can store a copy of smat and have mulliken_calc and fukui use that copy.

IgnacioJPickering:
I tried to do that today and it seems to involve quite a bit of work because I had to find which routines effectively use smat and which ones just have it passed on because they are calling lowdin_calc or mulliken_calc, or fukui down the line. If you agree with this I can try to implement it this week.

ramirezfranciscof:
Question: Why can't the subroutines eat the S-related matrix that exists in the calling procedure instead of having to store a copy of it?

Generally speaking I think one should always try as much as possible to have subroutines only modify things that are "outputs" (i.e.: no side effects), which only depend on the "inputs" (i.e.: no state-like dependencies). Of course this is not always possible, but I find it hard to understand why it wouldn't be in this case.

IgnacioJPickering:
@ramirezfranciscof I reverted the commit you asked about, that is why you can't find it anymore.

ramirezfranciscof:
[Ahhhh](http://i0.kym-cdn.com/entries/icons/original/000/022/138/reece.JPG), ok.

Ok, then I repeat my previous question: Why can't the subroutines eat the S-related matrix that exists in the calling procedure instead of having to store a copy of it?

lioamber/properties.f90
fedepedron: Please do try to avoid this. As a renamer myself, I admit it only adds a small contribution to the general chaos.

lioamber/fterm_biaspot.f90
```diff
@@ -1,6 +1,7 @@
 !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 !%% FTERM_BIASPOT.F90 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
 ! TO-DO: EXPLANATION.
+! This subroutine is used by the ehrenfest module only
```
fedepedron: Try to avoid making unrelated changes.

lioamber/init_lio.f90
```diff
@@ -44,7 +44,7 @@ subroutine lio_defaults()
 
 !   Names of files used for input and output.
     basis          = 'basis'       ; output             = 'output'      ;
-    fmulliken      = 'mulliken'    ; fcoord             = 'qm.xyz'      ;
+    fpopulation      = 'population_analysis'    ; fcoord             = 'qm.xyz'      ;
```
fedepedron: One tries to keep everything aligned.... On a sidenote, maybe these variables need to be moved to fileio module.

fedepedron:
So, what's the final status on this one? Are the changes @ramirezfranciscof already done?  (also, please rebase)

lioamber/fterm_biaspot.f90
IgnacioJPickering: Sorry about that, I commented that because the subroutine is the only other part of the program where sqsmat is used besides fockbias and lowdin, and I wanted it to be immediatly clear what it is used for, the name seems to imply that it is somehow related to the fockbias_subs and fockbias_data modules (what it does is related, but it is not used by them in any way), which made it confusing, I know its not immediatly related but I hope it doesn't hurt.

IgnacioJPickering:
I believe I changed what @ramirezfranciscof requested, I would like to run a transport dynamic with lowdin charges to see what the effect on performance actually is. I will try to do that during the weekend but I'm not sure I'll be able to, perhaps @charlyqchm can do it one of these days. @fedepedron will explain in detail how to rebase on monday =)

IgnacioJPickering:
Rebased and everything looks nice now, I still would like for charlyqchm to run some tests and see if he is happy with it

lioamber/properties.f90
```diff
@@ -100,27 +101,34 @@ end subroutine mulliken_calc
...
+    !sqsmat is built directly inside lowdin_calc since sqsmat is not really built anywhere
```
ramirezfranciscof: The second "built" should be "used", right? Ok, I would say this is information relevant for the PR (because you are justifying a change) but it really doesn't make sense to include this kind of comments in the code because if you see it for the first time it's kind of obvious that you calculate there because you use it then. There is no need to know or reference that it used to work differently.

fedepedron:
Status on this one?

gonzalodm:
i've compiled with gfortran and cuda=2 and works!!!. I tested this with
ehrenfest and BO dynamics.

2018-07-16 11:08 GMT-03:00 Federico N. Pedron <notifications@github.com>:

fedepedron:
I am sorry @IgnacioJPickering we kept this such a long time here, as of today Lowdin charges were fixed for OS (using what was in this PR) and the other problem it had (SQSM) was fixed by @ramirezfranciscof. I am closing this since there is no point in keeping it any longer.

PR #228, opened 2018-06-18
File IO - The Revenge of the Sith

fedepedron:
Some fixes and improvements in IO handling:
* Enabled rho-based restarts in LIO via the new variable rst_dens.
* Fixed TD restarts, which were broken somewhere in recent history.
* Added population analysis in TD via the new variable td_do_pop (lit. "td hace pop; pop para divertirse").

@eduarditoperez this should fix your issues with namelist reading.


ramirezfranciscof:
@fedepedron this is not working.

BTW: how does the rst_dens work exactly? Is it saving the density matrices for verlet and the fock matrices for magnus?

fedepedron:
Dammit makefile dependencies. And no, is to use rho restarts in SCF.

PR #231, opened 2018-07-10
Atoms rising from the Grave

fedepedron:
Added ghost atom functionality to LIO, in order to perform BSSE calculations and charge transfer calculations.

To use, simply set n_ghosts to the desired quantity, and ghots_atoms=x,y,z where x, y,z are the ghost atom indexes (max. 300).


charlyqchm:
Transport, DFTB and Open Shell seems to be working normally.

PR #240, opened 2018-08-15
Bugfix - the order of factors ALTERS the product.

fedepedron:
Fixed a bug in OS correlation calculation, in which some numbers ended in a "inf / inf = NaN" situation. Solved by altering the way operations were performed. Although calculations are similar in closed-shell, I could not find a system to reproduce this bug, so I left CS as it is.

Also, added a small check in OS exchange correlation for very small beta densities (in order to avoid further NaNs).

Left some NaN checks commented out on purpose, to turn them later in a proper debugging tool. I'll be merging this tomorrow.

### DIFF
--- a/g2g/pointxc/pbeOS_corr.h
+++ b/g2g/pointxc/pbeOS_corr.h
@@ -199,7 +199,7 @@ __host__ __device__ void pbeOS_corr(scalar_type rho, scalar_type rs,
   //      FACT1 = Q5*Q9+Q4*Q9*Q9
 
   scalar_type hB =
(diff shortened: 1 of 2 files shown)


ramirezfranciscof:
This is not a modification in "three lines of code".

fedepedron:
Three lines and an if-guard...

ramirezfranciscof:
Ok, I'll trust you on the things your taking out/putting in the if-guard, but may I ask why you removed the scalar_type declaration of the variables? Other than that it looks ok.

fedepedron:
> why you removed the scalar_type declaration of the variables?

I did not remove it, they are declared above. If not, those variables would've been declared only inside the if-guard.

PR #241, opened 2018-08-16
The ghost of a guess.

fedepedron:
I'll be merging this immediately, there was a compatibility bug between aufbau-like initial guess and ghost atoms, which have Iz = 0

### DIFF
--- a/lioamber/initial_guess.f90
+++ b/lioamber/initial_guess.f90
@@ -38,13 +38,13 @@
 !##############################################################################!
 module initial_guess_data
   integer :: initial_guess = 0
-  integer :: atomic_eec(54,3)
+  integer :: atomic_eec(0:54,3)
 
 contains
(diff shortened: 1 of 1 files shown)


lioamber/initial_guess.f90
```diff
@@ -38,13 +38,13 @@
 !##############################################################################!
 module initial_guess_data
   integer :: initial_guess = 0
-  integer :: atomic_eec(54,3)
+  integer :: atomic_eec(0:54,3)
```
ramirezfranciscof: Mmmm, have we become C programmers? why is this necessary?

fedepedron: Nothing to do with C. The thing is, ghost atoms have Iz=0; this is done this way to avoid checking "if Iz .gt. 0" later during the loops in the aufbau-like initial guess.

PR #243, opened 2018-08-28
Integral Refactor - Part III

fedepedron:
Reformated intfld, intsol, intSG and intsolg. 

Also, restored some old garchamod dependencies in order to easily move them to a new basis module later.


charlyqchm:
Everything seems to be working normally.

PR #245, opened 2018-09-04
Tamm dancoff

gonzalodm:
This PR performs a Linear Response calculate.
to use, add in the input file:
lresp = t,
nstates = x, number of excited states that you want.

This implementation only works with libxc and libint libraries and when you compiled with cuda = 0 for the moment.


segmentat10nfault:
Grosso!!!!

fedepedron:
Moved the base to new "develop" branch, in order to keep new developments there.

lioamber/drive.f90
```diff
@@ -222,18 +223,21 @@ SUBROUTINE drive(ng2,ngDyn,ngdDyn)
                       xnorm=sqrt(xnorm)
                       c(No,l)=ct(index)*xnorm
                       a(No,l)=at(index)
```
fedepedron: Instead of adding confusion in drive and init_lio, I'd prefer an "init_linear_response()" subroutine (or something like that) that allocates cbas, cbasx and initializes them. This way it can be called only when an LR calculation is desired.

ramirezfranciscof: This was later talked in person by @gonzalodm and @fedepedron and they decided to leave it this way here and fix it in the upcoming reformatting of drive.

PR #247, opened 2018-09-14
Minor improvements and bugfixes. 

fedepedron:
* When magnus restarts do not find fock, TD takes leapfrog steps.
* Fixed allocation issues for cx, ax, ncontx, nucx.
* Fixed Mulliken printing when called after SCF.
* Fixed Mulliken printing during Magnus leapfrog steps.


ramirezfranciscof:
Seems ok. In the future it might be better to have better format for restarts, maybe actually using titles and scanning the files for those rather than expecting things to be in a certain order.

charlyqchm:
It seems no problem. I tested DFTB and Transport with CUDA = 2 and INTEL = 2

PR #251, opened 2018-09-28
Linear Search for Rho

ramirezfranciscof:
Incorporation of the linear search algorithm to improve convergence of SCF in complicated cases. Further cleanup of the implementation would be required.


lioamber/fileio/output_others.f90
```diff
@@ -233,6 +233,7 @@ subroutine write_population(N, q0, q, pop, UID)
       write(UID,300)
       if (pop.eq.0) write(UID,301)
       if (pop.eq.1) write(UID,309)
+      if (pop.eq.2) write(UID,310)
```
fedepedron: I believe spin population analysis should be always printed; or at least whenever the system is open shell. This information is always useful.

lioamber/init_lio.f90
```diff
@@ -55,6 +56,7 @@ subroutine lio_defaults()
     GOLD           = 10.           ; omit_bas           = .false.       ;
     charge         = 0             ;
     fitting_set    = "DZVP Coulomb Fitting" ;
+    Rho_LS         = 0             ; P_oscilation_analisis = .false.
```
fedepedron: Shouldn't this be initialized in its own module? We agreed on not adding things to garcha mod or this initialization unless strictly necessary.

lioamber/input_read.f90
```diff
@@ -26,6 +27,17 @@ subroutine read_options(inputFile)
        write(*,*) 'File ', adjustl(inputFile), ' not found. Using defaults.'
     endif
 
+    if (Rho_LS .gt. 0 .and. (DIIS .or. hybrid_converg)) then
+      write(*,*)
```
fedepedron: Try to follow the standard I'm keeping with errors: write(*,'(A)') "  ERROR/WARNING: blablabla". This way the output file will return to a mess in no time.

lioamber/linear_search.f90
```diff
@@ -0,0 +1,253 @@
+!%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%!
+! Lineal search subroutines
+!
```
fedepedron: Shouldn't this subs be in "converger_subs"? After all, it IS a convergence algorithm. In addition, the name "linear_search" is confusing since it could refer to its geometry optimization variant. Or ANY linear search algorithm in the world for that matter.

lioamber/liomain.f90
```diff
@@ -17,13 +17,15 @@ subroutine liomain(E, dipxyz)
                           restart_freq, npas, sqsm, mulliken, lowdin, dipole, &
                           doing_ehrenfest, first_step, Eorbs, Eorbs_b, fukui, &
                           print_coeffs, steep,       MO_coef_at, MO_coef_at_b,&
-                          NUnp, NCO
+                          NUnp, NCO, spinpop, calc_propM
     use ecp_mod   , only: ecpmode, IzECP
     use ehrensubs , only: ehrendyn_main
     use fileio    , only: write_orbitals, write_orbitals_op
+    use steepest_descent, only: do_steep
```
fedepedron: In accordance with the other comment, I think "steepest descent" should be renamed to "geometry optimization" module.

lioamber/liomain.f90
```diff
@@ -48,7 +50,13 @@ subroutine liomain(E, dipxyz)
...
 
     ! Perform Mulliken and Lowdin analysis, get fukui functions and dipole.
-    if (MOD(npas, energy_freq).eq.0) then
+    calc_prop=.false.
+    if (MOD(npas, energy_freq).eq.0) calc_prop=.true.
+    if (calc_propM) calc_prop=.true.
+
+    if (calc_prop) then
+
+        if (mulliken.or.lowdin.or.spinpop) call do_population_analysis()
```
fedepedron: Same as before, spin population should be calculated every time Lowdin/Mulliken is called for an open shell system, no need to keep a separate keyword.

lioamber/liomods/garcha_mod.f
```diff
@@ -138,6 +138,22 @@ module garcha_mod
...
+
+! rho linear search
+      integer :: Rho_LS !linear sear in rho for convergention, 0 no LS, 1 LS only if energy increase, 2 full LS
+      double precision, dimension(:), allocatable :: rho_lambda1
+      double precision, dimension(:), allocatable :: rho_lambda0
+      double precision, dimension(:), allocatable :: rho_lambda1_alpha
+      double precision, dimension(:), allocatable :: rho_lambda0_alpha
+      double precision, dimension(:), allocatable :: rho_lambda1_betha
+      double precision, dimension(:), allocatable :: rho_lambda0_betha
+      double precision :: Pstepsize
```
fedepedron: Shouldn't this go to "converger_subs" instead of garcha?

test/18_Rho_LS/compile_options
```diff
@@ -0,0 +1 @@
+cuda=1 sm20=1 time=1
```
fedepedron: sm20? really?

lioamber/input_read.f90
ramirezfranciscof: TD line 732, your warning is with write(*,*) instead of write(*,'(A)'). Just sayin'...

Na, kidding. Making the change, but maybe we should consider having separate warning/error subroutine with centralized error messages in the future.

fedepedron:
And also, please add some documentation.

ramirezfranciscof:
Changes made, expanded the header of linear_search, now converger_ls. Compilation check passed, let me know if anything else needs changing.

fedepedron:
Thanks for the changes, I'll test it asap.

fedepedron:
> serius
LoL.  On a side topic (or main topic, really) I'm still unable to find the problems with spin density :S

ramirezfranciscof:
> On a side topic (or main topic, really) I'm still unable to find the problems with spin density :S

Is this broken in this PR? Was this working before?

Nick-Chaos:
Estimados, me gustaria que la incorporacion de mis rutinas al codigo en MALBECC se realice en su formato original.

PR #252, opened 2018-10-01
Setting the GPU level in runtime.

fedepedron:
Small PR here, now instead of recompiling over and over again with AINT_GPU_LEVEL, you just need to set "gpu_level=X" in the &lio namelist, where X is the desired GPU level (0 = no analytic integrals in GPU, and then scaling up to % = everything in GPU).

### DIFF
--- a/g2g/Makefile
+++ b/g2g/Makefile
@@ -48,10 +48,6 @@ else
 endif
 endif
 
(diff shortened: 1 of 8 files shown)


gonzalodm:
I tested this PR with cuda=2 intel=0 and with all gpu_level and it works
well.

PR #255, opened 2018-10-10
Smallfix

ramirezfranciscof:
* Added slightly better keyword to print spin (this still needs improvement).
 * Fixed spin calculation (needed a call to fixrho)
 * Deleted old version of SCF saved in ehrenfest subs from before we knew how to properly work with git.


lioamber/liomain.f90
```diff
@@ -74,6 +79,9 @@ subroutine liomain(E, dipxyz)
              call write_orbitals(M, NCO, Eorbs, MO_coef_at, 29)
           endif
         endif
+    else if (spinpop) then
```
fedepedron: Didnt we agree on NOT having a specific keyword? This should be printed when mulliken = t.

ramirezfranciscof: Yes, that's right. I did it this other way because I wasn't exactly sure of the logic it used to decide whether to call do_population_analysis or not in liomain, so in order to not mess up with that I included an extra keyword to force the call. I'm pretty sure that it would still print the spin if your manage to trigger that "if (calc_prop) then", but apparently mulliken = t was not doing it...

If you can tell me how the code is supposed to decide to call the population analysis I can modify that; maybe I'm just misunderstanding something, but as it stands right now it doesn't seem to be working properly.

ramirezfranciscof:
UPDATING DESCRIPTION:

* Bugfix: Fixed spin calculation (needed a call to fixrho). Also,
   now it is only called when mulliken is requested.
   (because it is the "mulliken" version of spin population)

 * Cleanup: Deleted old version of SCF saved in ehrenfest subs from
   before we knew how to properly work with git.

 * Cleanup: Removed spinpop variable that was not being used.

 * Restruct: sent the npas counter add to liomain instead of scf.

PR #265, opened 2018-11-20
Cleanup - second issue

fedepedron:
Moved tmpaux, test_subs and liosubs_dens to a new SCF_aux module containing auxiliary subroutines for SCF.
Moved some maskrmm subroutines into ehrensubs and deleted maskrmm.
Also removed unused ehrenfest files.


lioamber/SCF.f90
```diff
@@ -55,9 +55,9 @@ subroutine SCF(E)
    use mask_ecp      , only: ECP_init, ECP_fock, ECP_energy
    use typedef_sop   , only: sop              ! Testing SOP
    use fockbias_subs , only: fockbias_loads, fockbias_setmat, fockbias_apply
-   use tmpaux_SCF    , only: neighbor_list_2e
+   use SCF_aux       , only: neighbour_list_2e, seek_nan, standard_coefs, &
+                             messup_densmat, fix_densmat
```
ramirezfranciscof: I have no problem with SCF_aux being a new "temporal" module, but just to note that seek_nan, messup_densmat and fix_densmat are general purpose functions that could be used elsewhere (actually, I'm surprised they are currently not).

PR #268, opened 2019-01-16
Cleanup 4: Basis and others

fedepedron:
First PR of the year! yay!

Now seriously, I finally moved everything to the new basis_data module, removing the related variables from garcha_mod and getting a better encapsulation. Also edited drive removing the operations already done by the basis module.

In addition in addition, general_module has been deleted, elec.f was merged into cubegen module, and lio_finalize was moved to f90, thus leaving **no ".f" files** in the main lioamber folder. Cheers!


charlyqchm:
Transport seems to be working normally. Tested with CUDA=2 and INTEL=2.

PR #270, opened 2019-01-23
Lioman

ramirezfranciscof:
Modifications for the lio manual.

This is a work in process, I'm posting the PR now because it might be of interest for the rest and I may soon be requesting contributions from other members to fill the different parts of it. That means this is not ready to be merged yet; it is, however, open to opinions and suggestions.


fedepedron:
I am going to merge this so future upgrades work from RC and not from this PR.

PR #278, opened 2019-03-28
New Tests

gonzalodm:
- This PR has a small program that compile LIO with differents options (CPU/GPU, INTEL/GNU, FLOAT/DOUBLE precision) and runs all tests (this doesn't include tests for amber or libxc yet)
- I deleted some tests


fedepedron:
Good PR, a few comments:

- Maybe unit_test and integration_test folders should be kept in a separate "libxc_tests" folder; and then run them if LIO has been compiled with LIBXC.

- I would remove for now 6_GeomOpt, 10_TransDens, 11_RhoLS,  12_RhoLS2 and testLS, since those features are only in the develop branch.

- Maybe add a Makefile or a run script in the main /tests folders? Seems somehow counter-intuitive to enter test_engine in order to run everything.

fedepedron:
Awesome. One last thing, ECP test has an issue with the LIOHOME, it is set to /opt/progs/LIOs/NewTest and not to the current directory.

PR #281, opened 2019-04-03
Precision issues

fedepedron:
Fixed all precision issues. I think. Hopefully. At least, fixed those I found.


fedepedron:
@charlyqchm, @gonzalodm please test.

charlyqchm:
Everything seems to be working correctly. Tested SCF, TD and TB-DFT with cuda=2 and intel=2.

PR #290, opened 2019-04-17
Removed all unused variables.

fedepedron:
Now there are no "unused variable" warnings except for those triggered by compilation options.


fedepedron:
Maybe this should be done with develop, at first I thought it was a good idea but it may add more confusion into rc.

PR #295, opened 2019-05-30
Datatypes: CuMat data structure and complex number precision

fedepedron:
* Added a FORTRAN header containing compiler pragmas for complex precision, standardising the type "TDCOMPLEX" from now on.
* Created CUMAT data structure (cumat_r, cumat_x for real and complex respectively). This allows us to encapsulate cublas-related issues. The procedures included do not span every possible thing you would want to do with those, but nevertheless it should be easy to expand as much as needed.

This drastically reduced the amount of code needed for TD and Magnus, so me is happy.


charlyqchm:
Allocation issue presented in the subroutine td_overlap_diag. X_mat and Y_mat are being allocated with the same dimension than X_min and Y_min (M), instead of M_f which includes the TBDFT case. Check the dimensions of the loops too.

fedepedron:
> Allocation issue presented in the subroutine td_overlap_diag. X_mat and Y_mat are being allocated with the same dimension than X_min and Y_min (M), instead of M_f which includes the TBDFT case. Check the dimensions of the loops too.

Can you fix them? I have no way to tell if I fixed it if I don't have a TB test .

PR #297, opened 2019-07-01
Constrained DFT and Becke charges.

fedepedron:
Added a CDFT implementation using a Becke partitioning scheme.
Some changes were needed in liomain to better accommodate these things
Option input is made via a new {CDFT} block in the input file, please refer to the test provided as an example.


charlyqchm:
Everything seems to be working normally. Checked TD-TBDFT with GPU and INTEL.

fedepedron:
@charlyqchm have you actually tried the CDFT module? Just out of curiosity.

PR #300, opened 2019-07-25
TB Fermi energy calculation - DFT charge calibration

charlyqchm:
In this pull request, a new subroutine was implemented for calculating the Fermi energy of the TB part using as control parameter the total charge in the DFT part. This calculation is iterative, in each step a new SCF calculation runs with a Fermi energy modified and comparing  the DFT charge with an input reference.


fedepedron:
Good PR! 
Unpopular opinion: may be we should separate TB variables in a new &tbdft namelist?

PR #305, opened 2019-08-29
Modifications for actual hybrid version

Nick-Chaos:
-included lio version check for hybrid-lio calculations
-modified force units for hybrid
-included spin check
-moved fa and fmm to allocatable arrays


fedepedron:
Everything seems fine. Since this is a very small change, I'll be merging it ASAP.

PR #314, opened 2019-10-07
DFT-D3

fedepedron:
Added Grimme's DFT-D3 corrections to energies and forces. 
Corrections are not particularly big (for the test with 2 TRPs, the energy correction is about ~2kcal/mol) but it is done and it may be useful in bigger systems.


fedepedron:
@gonzalodm @charlyqchm a little code review maybe?

PR #325, opened 2020-02-03
pbe0 works in TD

gonzalodm:
Add pbe0 functional to TD calculation.


fedepedron:
Code seems fine and everything is working. Maybe after this we should start masking PBE0 usage better  (the thing added in propagators seems carbon-copied from the one in SCF).

PR #326, opened 2020-02-05
Excited prop

gonzalodm:
Add excited state forces calculation and non-adiabatic coupling vectors.


fedepedron:
This is madness. But everything works fine! Also, all tests passed.

PR #336, opened 2020-03-30
TSH calculations

gonzalodm:
LIO performs TSH dynamic only when is called by HYB. In order to use it, you have to compile Hybrid and LIO with tsh=1 option.


fedepedron:
Never less than a thousand lines huh?
I have two questions:
a) Why only hybrid?
b) Is it posible to move the "decide if hopping" thing from scf_hyb to liomain?

lioamber/init_lio.f90
```diff
@@ -434,7 +436,7 @@ subroutine init_lio_hybrid(version_check, hyb_natom, mm_natom, chargein, iza, sp
     character(len=20)   :: inputFile
     integer, intent(in) :: chargein   !total charge of QM system
     integer, dimension(hyb_natom), intent(in) :: iza  !array of charges of all QM/MM atoms
-    LIODBLE, intent(in) :: spin !number of unpaired electrons
+    LIODBLE, intent(in) :: dt, spin !number of unpaired electrons, dt = time step in femtosec
     integer :: Nunp_aux !auxiliar
 
     if (version_check.ne.1) Stop 'LIO version is not compatible with hybrid'
```
Nick-Chaos: Para no perder la compatibilidad entre las versiones actuales de lio y hybrid es recomendable cambiar el valor q compara contra version_check cada vez que se modifican los argumentos de las subrutinas init_lio_hybrid y SCF_hyb. Obiamente el cambio tiene q estar acompañado por un PR en hybrid cambiando los argumentos.

gonzalodm:
The final decision about the hopp is not in scf_hyb!!!, is in TSH_hyb. I
need hybrid because I need to call LIO from MM program after velocities
actualization and before the position actualization. In Hybrid is easier to
do this than amber. In the next days, I will submit a PR in Hybrid with the
corresponding modifications.

PR #341, opened 2020-04-28
completed list of masses

Nick-Chaos:
included masses upto Xe


fedepedron:
Trivial change, merging.

PR #355, opened 2020-07-28
Corrections and improvements of TBDFT and DOS subs

charlyqchm:
The next changes have been included:
- Some bugs have been fixed in TBDFT subroutines.
- Now spin current and spin population are printed
  during TBDFT open shell calculations.
- Alfa and Beta DOS and PDOS are now printed
  separated.

### DIFF
--- a/lioamber/dos_subs.f90
+++ b/lioamber/dos_subs.f90
@@ -91,7 +91,10 @@ subroutine build_PDOS(coef_mat, overlap, M, M_total, Nuc)
          enddo
          enddo
       end do
+
    end if
+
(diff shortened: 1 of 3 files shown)


lioamber/dos_subs.f90
```diff
@@ -73,6 +70,8 @@ subroutine build_PDOS(coef_mat, overlap, M, M_total, Nuc)
       enddo
 
    if (pdos_allb) then
+      pdos_b = 0.0d0
+      if (.not. allocated(index_b)) allocate(index_b(pdos_nbases))
```
fedepedron, line 74: index_b is never deallocated.

PR #356, opened 2020-09-15
Second lr

gonzalodm:
Add two methods for excited states spectra (Fosc and SecondLR)


fedepedron:
Everything seems fine (although no open shell... yet). Please do add a proper deallocation for tdipsave. somewhere.

PR #363, opened 2021-11-14
Add a Github "build all" and "build and test" actions to replace Travis

manuelF:
Travis is no longer the preferred tool (many changes around open source availability).

Github actions are native to Github and run fast. There is also a huge ecosystem where they other things such as linters, coming in another PR.

Still no GPU support [1]. However this PR adds also Build in GPU configurations (and full double) so it at least covers the cases where it works.

Link to a full execution of these actions shows the right things [2][3]

[1] https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
[2] https://github.com/manuelF/lio/actions/runs/1458487902
[3] https://github.com/manuelF/lio/actions/runs/1458742553

### DIFF
--- a/.github/workflows/build-all.yaml
+++ b/.github/workflows/build-all.yaml
@@ -1,0 +1,28 @@
+# This action compiles only the different variants of CPU and GPU.
+#
+# It does not run tests since the workers are CPU-only Virtual Machines.
+# To enable GPU tests it would have to build a GPU version and run it on
+# an external Github worker in some other cloud.
+
+name: Compile-only CPU and CPU-GPU variants
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: 'Build with ${{ matrix.cuda_option }} ${{ matrix.precision }} ${{ matrix.analytics }}'
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        cuda_option: ['cuda=0', 'cuda=1']
+        precision: ['precision=0', 'precision=1']
+        analytics: ['analytics=0', 'analytics=2']
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: get packages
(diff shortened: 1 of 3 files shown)


fedepedron:
Nice update!

So even though it does not allow for GPU runs you _do_ have CUDA libraries available to at least try and compile. Which, IIRC, was not the case with travis (at least in the beginning), so that's definitely an upgrade.

A couple of things to note though:
  - cpu is no longer a compile option since it was basically redundant.
  - full_double has been replaced with precision=1. Although full_double is still available, it's currently deprecated.
  - maybe for compilation-only tests it could be interesting to add the analytics=2 flag, which basically enables warnings and debug information.

manuelF:
Thanks Fede, applied the changes and now builds the cuda 0/1, analytics 0/2, precision 0/1 configs.

Though I am not super sure about the precision and CPU flags:
https://github.com/MALBECC/lio/blob/73070f6ffd1fc991c5652ecf943cd249c35a8df2/g2g/Makefile#L72
https://github.com/MALBECC/lio/blob/73070f6ffd1fc991c5652ecf943cd249c35a8df2/g2g/Makefile.cuda#L109

These two makefile changes might still have some impact. Any idea what should be done?

fedepedron:
In the case of CPU, that line should probably removed, I'm not sure things would work properly with cpu=0, but we may keep it as it is just for testing purposes.

In the case of full_doble, see Makefile.translate within the same folder:

https://github.com/MALBECC/lio/blob/73070f6ffd1fc991c5652ecf943cd249c35a8df2/g2g//Makefile.translate#L52

When precision=1, full_double is also enabled. The problem is that using full_double will not enable double precision in other parts of the FORTRAN code base.

fedepedron (approved):

Issue #109, opened 2017-07-12
To add to a C/CPP/CUDA conventions / best practices


fedepedron:
(I'm @ramirezfranciscof )

Things to add to the guide on how 

(1) Add headers in the .cpp files when possible, in the .h when there is no choice (ex: when declaring types that are defined in another .h). [source](https://stackoverflow.com/questions/16821763/where-should-i-include-header-files-in-the-cpp-or-in-the-h-file)

(2) Includes always outside of namespaces.

To be continued...

fedepedron:
¿Perhaps implementing the #ifndef / #define / #endif guard inside headers?

fedepedron:
(3) Families of subroutines inside namespaces, classes are for objects. [Source](https://stackoverflow.com/questions/1434937/namespace-functions-versus-static-methods-on-a-class)

(4) When classes are inside a namespace, the definition of is methods is as shown in [this link](https://stackoverflow.com/questions/8681714/correct-way-to-define-c-namespace-methods-in-cpp-file). It seems to usually be a good a idea to put classes inside namespaces ([source](https://stackoverflow.com/questions/28135266/why-to-use-classes-inside-namespace)).

fedepedron:
(5) Resolve issue of where to define templates (.h or .cpp) and where to specialize them (see [first answer](https://stackoverflow.com/questions/972152/how-to-create-a-template-function-within-a-class-c))

fedepedron:
(6) First write #includes<banana>, then #includes "localbanana". And sort them alphabetically.

manuelF:
#160 is the first step, where everything is formatted. After this, all the improvements make more sense on a standandarized codebase.

PR #365, opened 2021-11-15
Clang-format all files (.cpp, .h, .cu)

manuelF:
Just like in #109, run the tool with the changes since https://github.com/MALBECC/lio/pull/160

```
cd g2g
find . | grep -E "\.(h|cpp|cu)$" | xargs clang-format-10 -i -style=file
```
This PR might bring a lot of conflicts to others, but the solution is to run the command above *before you merge*. This should cause your diffs to be zero.

Cuda changes are now correctly formatted (after clang added support https://reviews.llvm.org/D42589) but there are a *lot*. They are all auto-generated but if people reformat-before-merge before there should be no problem.

After fixing the formatting, `cpp-lint` shows the following issues:
https://github.com/manuelF/lio/runs/4218078570?check_suite_focus=true

Likely not all of them are worth fixing, so the tool can be configured to ignore certain kind of issues, like copyright. 
(see https://android.googlesource.com/platform/art/+/master/CPPLINT.cfg for example)

I will configure it independently in a separate PR.

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -28,7 +28,9 @@ jobs:
           # We enable explicitly YAML so it doesn't check for anything else.
           # Only YAML should be .github/workflows.
           VALIDATE_YAML: true
-          # TODO: Enable this when codebase is clang-format compliant.
+          # Validates C++ (.cpp, .h and .cu) match the standard.
+          VALIDATE_CLANG_FORMAT: true
+          # TODO: Enable this when codebase is cpplint compliant.
           # VALIDATE_CPP: true
           # TODO: Enable this when codebase is autopep8 compliant.
           # VALIDATE_PYTHON: true
--- a/g2g/analytic_integral/aint_init.cpp
+++ b/g2g/analytic_integral/aint_init.cpp
@@ -12,11 +12,11 @@
 #include "qmmm_integral.h"
 #include "coulomb_integral.h"
 
+using std::boolalpha;
 using std::cout;
 using std::endl;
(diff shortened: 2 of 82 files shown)


fedepedron:
Hmm this one might need more consensus. Opinions?
@nanolebrero @gonzalodm @ramirezfranciscof @charlyqchm

ramirezfranciscof:
What is this? Did I step on a time machine?

My opinion is: lets refactor the codebase into python while the argentinians are sleeping.

(On a more serious note, I'm a little rusty on this so a couple of questions: I only see changes on the sourcefiles, so the linter was already configured in the `.git` files? I don't remember anything triggering on commits. Also, is there an automatic formatter for these files?)

manuelF:
Hi, all of this is automatically generated via the compiler (clang), that enables a standard formatting in the codebase. This was done in https://github.com/MALBECC/lio/pull/160 but it didn't include cuda files then, as the formatter did not support it correctly. Now it does.

The automatic formatter-checker will now be enabled in order then to not have any other regressions: https://github.com/MALBECC/lio/pull/364 (only when trying to merge onto develop and master, as to allow people to freely experiment and not getting any complaints)

Basically this just standarizes the entire codebase (C++/Cuda for now), old and new, into a single unified style.

manuelF:
Checked that after this PR all the code is clang-format compliant. New or modified C++ and CUDA code will be checked against style to prevent regressions (only on `master` and `develop` branches).

manuelF:
Abandoning this PR as this will probably never get merged.

PR #366, opened 2021-11-18
Reformat python to match black standard

manuelF:
This PR reformats all the code to match the pep8 python standard using `black` tool, and enables the linter to prevent further regressions.
 
Only non-automated change I did was add `r"regex"` (raw-strings) which is the recommended and correct way of using it in Python3 (See python3 re documentation: https://docs.python.org/3/library/re.html#raw-string-notation)

Command ran:
 
`black .`

https://github.com/psf/black (`apt install black`)

Currently ran

```
$ black --version
black, version 19.3b0
```

Next PR will contain actual code fixes with some type hints.


manuelF:
Abandoning this PR as it will probably never get merged.

